PR #8, opened 2018-03-05
Use loadable-components

mattgrill:
- use `loadable-components`
- don't send down parts of the app that we don't need.
- currently only on entire page components, but can probably break this up further.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "private": true,
   "dependencies": {
     "emotion": "^9.0.2",
+    "loadable-components": "^1.1.1",
     "react": "^16.2.0",
     "react-dom": "^16.2.0",
     "react-router-dom": "^4.2.2"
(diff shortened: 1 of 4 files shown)


justafish:
I'm not against doing this in future, but I think it's a premature optimisation to do now. (I'm guessing the extra network requests will actually slow it down given the current size of our app 😂)

dawehner:
loadable-components sounds quite nice! I think having something like this would be really valueable on the longrun. At this moment in time I have a hard time thinking whether X is a right solution. For me it seems like we don't even know the problem all over the place.

@matt Would ```loadable-components``` work when we pass along a JS path/URL containing a component?

dawehner:
@drpal @justafish @dawehner talked about this and we agreed to leave it for now. We will come back to module component loading later, which will be "fun" or fun.

Issue #7, opened 2018-03-04
Stub out the permissions page


justafish:
## API Endpoints Needed
- [list of permissions](https://www.drupal.org/project/drupal/issues/2947605)
- Roles and Permissions (config entities) are available from JSON API

We're already going to be on the same domain as Drupal, meaning we can use Cookie authentication, so we can use the browser's fetch API directly for this.

## Components Needed
- [Table (with sticky header)](https://github.com/jsdrupal/drupal-react-components/tree/master/src/components/table)
- [Checkbox](https://github.com/jsdrupal/drupal-react-components/tree/master/src/components/checkbox)

## Submit Function
I'm not sure if we can write the config entities through JSON API yet, so for now let's pretend it's making a successful request with an artificial delay

## Permissions Page
Component skeleton is [here](https://github.com/jsdrupal/drupal-admin-ui/blob/master/src/components/05_pages/Permissions/Permissions.jsx), with ability to add CSS 
See also https://github.com/jsdrupal/drupal-admin-ui/tree/master/src/styles for global styles

## Jest Test
Can be added as an additional file in https://github.com/jsdrupal/drupal-admin-ui/tree/master/src/components/05_pages/Permissions

## Bonus Round
It would be great to add Storybook to this project and add a .story file into https://github.com/jsdrupal/drupal-admin-ui/blob/master/src/components/05_pages/Permissions

justafish:
Should we start importing the components into this repo so it's all in one place?

dawehner:
@all Do you consider this stub out to be successful?

tedbow:
@dawehner it is probably good for a stub. 

It would be good if there was message that the permissions were saved but maybe we could do that in follow up?

dawehner:
Yeah let's open follow ups.

dawehner:
I opened up another follow for some other details.

dawehner:
I'd say we are good on that one.

PR #9, opened 2018-03-06
#7 Permissions

mattgrill:
### Setup

- Enable [JSON API](https://www.drupal.org/project/jsonapi) and [Admin UI support](https://github.com/jsdrupal/drupal-admin-ui-support)modules.

### Permissions

- [x] Request and render permissions list from Drupal.
- [x] Need to refine json from drupal for permissions list, _why are we sending html?_
- [x] Need to figure out how to reference current permission to user role data.
- [x] Live filtering?
- [ ] Styles

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "private": true,
   "dependencies": {
     "emotion": "^9.0.2",
+    "interweave": "^8.0.2",
     "makecancelable": "^1.0.0",
     "prop-types": "^15.6.1",
     "react": "^16.2.0",
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "private": true,
   "dependencies": {
     "emotion": "^9.0.2",
+    "makecancelable": "^1.0.0",
     "prop-types": "^15.6.1",
     "react": "^16.2.0",
     "react-dom": "^16.2.0",
--- a/src/components/05_pages/Permissions/Permissions.js
+++ b/src/components/05_pages/Permissions/Permissions.js
@@ -36,16 +36,12 @@ const Permissions = class Permissions extends Component {
             roles: roles.sort((a, b) => {
               if (a.attributes.is_admin && b.attributes.is_admin) {
                 return a.attributes.id - b.attributes.id;
-              }
-              else if (a.attributes.is_admin) {
+              } else if (a.attributes.is_admin) {
                 return 1;
-              }
-              else if (b.attributes.is_admin) {
+              } else if (b.attributes.is_admin) {
                 return -1;
               }
-              else {
-                return a.attributes.id - b.attributes.id;
-              }
+              return a.attributes.id - b.attributes.id;
             }),
             loaded: true,
           }),
@@ -56,18 +52,16 @@ const Permissions = class Permissions extends Component {
   componentWillUnmount() {
     this.cancelFetch();
   }
-  groupPermissions = permissions =>
-    Object.entries(
-      permissions.reduce((acc, cur) => {
-        acc[cur.provider] = acc[cur.provider] || [];
-        acc[cur.provider].push(cur);
-        return acc;
-      }, {}),
-    );
+  onPermissionCheck = (roleName, permission) => {
+    this.setState(prevState => ({
+      changedRoles: [...new Set(prevState.changedRoles).add(roleName).values()],
+      roles: this.togglePermission(permission, roleName, prevState.roles),
+    }));
+  };
 
   togglePermission = (permission, roleName, roles) => {
-    let roleIndex = roles.map(role => role.attributes.id).indexOf(roleName);
-    let role = roles[roleIndex];
+    const roleIndex = roles.map(role => role.attributes.id).indexOf(roleName);
+    const role = roles[roleIndex];
     const index = role.attributes.permissions.indexOf(permission);
     if (index !== -1) {
       role.attributes.permissions.splice(index, 1);
@@ -78,16 +72,14 @@ const Permissions = class Permissions extends Component {
     return roles;
   };
 
-  onPermissionCheck = (roleName, permission) => {
-    this.setState((prevState, props) => {
-      return {
-        changedRoles: [
-          ...new Set(prevState.changedRoles).add(roleName).values(),
-        ],
-        roles: this.togglePermission(permission, roleName, prevState.roles),
-      };
-    });
-  };
+  groupPermissions = permissions =>
+    Object.entries(
+      permissions.reduce((acc, cur) => {
+        acc[cur.provider] = acc[cur.provider] || [];
+        acc[cur.provider].push(cur);
+        return acc;
+      }, {}),
+    );
 
   createTableRows = (groupedPermissions, roles) =>
     [].concat(
@@ -108,7 +100,7 @@ const Permissions = class Permissions extends Component {
                 ) : (
                   <input
                     type="checkbox"
-                    onChange={e =>
+                    onChange={() =>
                       this.onPermissionCheck(attributes.id, permission.id)
                     }
                     checked={attributes.permissions.includes(permission.id)}
@@ -124,8 +116,8 @@ const Permissions = class Permissions extends Component {
     this.setState(prevState => ({
       ...prevState,
       renderablePermissions: prevState.rawPermissions.filter(
-        ({ title, description, provider, provider_label }) =>
-          `${title}${description}${provider}${provider_label}`.includes(input),
+        ({ title, description, provider, provider_label: providerLabel }) =>
+          `${title}${description}${provider}${providerLabel}`.includes(input),
       ),
     }));
   };
--- a/src/components/05_pages/Permissions/Permissions.js
+++ b/src/components/05_pages/Permissions/Permissions.js
@@ -1,8 +1,8 @@
 import React, { Component, Fragment } from 'react';
 import makeCancelable from 'makecancelable';
+import { Markup } from 'interweave';
 
 import Loading from '../../Helpers/Loading';
-
 import { Table, TableBody, TableHeaderSimple } from '../../UI';
 
 const Permissions = class Permissions extends Component {
@@ -60,7 +60,7 @@ const Permissions = class Permissions extends Component {
         ...permissions.map(permission => ({
           key: `permissionGroup-${permissionGroupName}-${permission.title}`,
           tds: [
-            permission.title,
+            <Markup content={permission.title} />,
             ...roles.map(
               ({ attributes }) =>
                 attributes.is_admin && attributes.id === 'administrator' ? (
--- a/src/components/05_pages/Permissions/Permissions.js
+++ b/src/components/05_pages/Permissions/Permissions.js
@@ -72,43 +72,48 @@ const Permissions = class Permissions extends Component {
   groupPermissions = permissions =>
     Object.entries(
       permissions.reduce((acc, cur) => {
-        acc[cur.provider] = acc[cur.provider] || [];
-        acc[cur.provider].push(cur);
+        acc[cur.provider] = acc[cur.provider] || {
+          providerLabel: cur.provider_label,
+          permissions: [],
+        };
+        acc[cur.provider].permissions.push(cur);
         return acc;
       }, {}),
     );
   createTableRows = (groupedPermissions, roles) =>
     [].concat(
-      ...groupedPermissions.map(([permissionGroupName, permissions]) => [
-        {
-          key: `permissionGroup-${permissionGroupName}`,
-          colspan: roles.length + 1,
-          tds: [[`td-${permissionGroupName}`, <b>{permissionGroupName}</b>]],
-        },
-        ...permissions.map(permission => ({
-          key: `permissionGroup-${permissionGroupName}-${permission.title}`,
-          tds: [
-            [
-              `td-${permissionGroupName}-${permission.title}`,
-              <Markup content={permission.title} />,
+      ...groupedPermissions.map(
+        ([providerMachineName, { providerLabel, permissions }]) => [
+          {
+            key: `permissionGroup-${providerMachineName}`,
+            colspan: roles.length + 1,
+            tds: [[`td-${providerMachineName}`, <b>{providerLabel}</b>]],
+          },
+          ...permissions.map(permission => ({
+            key: `permissionGroup-${providerMachineName}-${permission.title}`,
+            tds: [
+              [
+                `td-${providerMachineName}-${permission.title}`,
+                <Markup content={permission.title} />,
+              ],
+              ...roles.map(({ attributes }, index) => [
+                `td-${providerMachineName}-${permission.title}-${index}-cb`,
+                attributes.is_admin && attributes.id === 'administrator' ? (
+                  <input type="checkbox" checked />
+                ) : (
+                  <input
+                    type="checkbox"
+                    onChange={() =>
+                      this.onPermissionCheck(attributes.id, permission.id)
+                    }
+                    checked={attributes.permissions.includes(permission.id)}
+                  />
+                ),
+              ]),
             ],
-            ...roles.map(({ attributes }, index) => [
-              `td-${permissionGroupName}-${permission.title}-${index}-cb`,
-              attributes.is_admin && attributes.id === 'administrator' ? (
-                <input type="checkbox" checked />
-              ) : (
-                <input
-                  type="checkbox"
-                  onChange={() =>
-                    this.onPermissionCheck(attributes.id, permission.id)
-                  }
-                  checked={attributes.permissions.includes(permission.id)}
-                />
-              ),
-            ]),
-          ],
-        })),
-      ]),
+          })),
+        ],
+      ),
     );
   handleKeyPress = event => {
     const input = event.target.value.toLowerCase();
--- a/src/components/05_pages/Permissions/Permissions.js
+++ b/src/components/05_pages/Permissions/Permissions.js
@@ -3,7 +3,7 @@ import makeCancelable from 'makecancelable';
 import { Markup } from 'interweave';
 
 import Loading from '../../Helpers/Loading';
-import { Table, TBody, THead } from '../../UI/Table';
+import { Table, TBody, THead } from '../../UI/table';
 
 const Permissions = class Permissions extends Component {
   state = {
--- a/src/components/05_pages/Permissions/Permissions.js
+++ b/src/components/05_pages/Permissions/Permissions.js
@@ -3,11 +3,13 @@ import makeCancelable from 'makecancelable';
 import { Markup } from 'interweave';
 
 import Loading from '../../Helpers/Loading';
-import { Table, TableBody, TableHeaderSimple } from '../../UI';
+import { Table, TableBody, TableHeaderSimple } from '../../UI/Table';
 
 const Permissions = class Permissions extends Component {
   state = {
     loaded: false,
+    rawPermissions: [],
+    renderablePermissions: [],
   };
   componentDidMount() {
     this.cancelFetch = makeCancelable(
@@ -21,6 +23,7 @@ const Permissions = class Permissions extends Component {
           `${
             process.env.REACT_APP_DRUPAL_BASE_URL
           }/jsonapi/user_role/user_role`,
+          { headers: { Accept: 'application/vnd.api+json' } },
         ).then(res => res.json()),
       ])
         .then(([permissions, { data: roles }]) =>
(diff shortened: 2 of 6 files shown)


src/components/05_pages/Permissions/Permissions.jsx
```diff
@@ -1,12 +1,63 @@
...
+      .then(permissions => {
+        const groupedPermissions = {};
+        Object.keys(permissions).forEach(key => {
+          groupedPermissions[permissions[key].provider] =
+            groupedPermissions[permissions[key].provider] || [];
+          groupedPermissions[permissions[key].provider].push(permissions[key]);
+        });
+        return { groupedPermissions, loaded: true };
+      })
+      .then(data => this.setState({ ...data }))
```
dawehner: Why can we not use ```.then(data => this.setState)```?

src/components/05_pages/Permissions/Permissions.jsx
```diff
@@ -1,12 +1,63 @@
...
+      .catch(console.log);
```
dawehner: Let's console.error that?

.eslintrc.json
```diff
@@ -0,0 +1,17 @@
+{
+  "extends": [
+    "eslint-config-airbnb",
+    "plugin:prettier/recommended"
```
dawehner: Nice!

src/components/05_pages/Permissions/Permissions.jsx
```diff
@@ -1,12 +1,63 @@
...
+        ) : (
+          <table>
+            <thead />
+            <tbody>
+              {Object.keys(this.state.groupedPermissions).map(
+                permissionGroupName => (
+                  <Fragment key={`fragment-${permissionGroupName}`}>
+                    <tr key={`permissionGroup-${permissionGroupName}`}>
+                      <td>{permissionGroupName}</td>
+                    </tr>
```
dawehner: I just love how prettier ensures things aren't too dense, honestly!

dawehner:
>  Need to refine json from drupal for permissions list, why are we sending html?

dawehner:
> Need to figure out how to reference current permission to user role data.
When you access the user roles they will each have a list of permissions they have

> Sortable?
Do you think this is super useful? I can't really think of a usecase for that.

> Live filtering?
Yes!

src/components/05_pages/Permissions/Permissions.jsx
```diff
@@ -1,12 +1,63 @@
...
-  </div>
-);
+const drupalOrigin = 'http://localhost:9998';
+
+const Permissions = class Permissions extends Component {
+  state = {
+    loaded: false,
+  };
+  componentDidMount() {
+    fetch(`${drupalOrigin}/permissions?_format=json`)
```
dawehner: I think we should group the permissions when we render. Otherwise live filtering permissions would be a bit more pain. Thoughts?

.eslintignore
```diff
@@ -0,0 +1 @@
+build/**/*.js
```
justafish: did https://github.com/jsdrupal/drupal-admin-ui/blob/master/.gitignore#L10 not cover this?

.eslintrc.json
justafish: 🤓

.eslintrc.json
```diff
@@ -0,0 +1,17 @@
...
+  "parser": "babel-eslint",
+  "root": true,
+  "env": {
+    "browser": true,
+    "es6": true
+  },
+  "rules": {
+    "no-param-reassign": [0],
+    "no-underscore-dangle": [0],
+    "react/jsx-filename-extension": [1, { "extensions": [".js", ".jsx"] }]
```
justafish: what's the advantage of doing this?

.prettierrc.json
```diff
@@ -0,0 +1,6 @@
+{
+  "printWidth": 80,
+  "semi": true,
+  "singleQuote": true,
+  "trailingComma": "all"
```
justafish: 💃

package.json
```diff
@@ -9,9 +9,19 @@
     "react-router-dom": "^4.2.2"
   },
   "devDependencies": {
+    "eslint": "^4.9.0",
```
justafish: Aren't these already included with CRA?

src/App.jsx
```diff
@@ -20,7 +16,7 @@ class NoMatch extends Component {
   render() {
     return null;
   }
-};
```
justafish: Was this a prettier fix?

src/components/05_pages/Permissions/Permissions.jsx
```diff
@@ -1,12 +1,63 @@
...
+import React, { Fragment, Component } from 'react';
 import { css } from 'emotion';
 
-const Permissions = () => (
-  <div>
-    <h1 className={styles.title}>Permissions</h1>
-    <p>This will be the permissions page.</p>
-  </div>
-);
+const drupalOrigin = 'http://localhost:9998';
```
justafish: let's add this in with dotenv (already included in CRA)

src/components/05_pages/Permissions/Permissions.jsx
justafish: Maybe we could route it through winston and have `logger.error` here instead?
https://www.npmjs.com/package/winston

src/components/05_pages/Permissions/Permissions.jsx
```diff
@@ -1,12 +1,63 @@
...
+        return { groupedPermissions, loaded: true };
+      })
+      .then(data => this.setState({ ...data }))
+      .catch(console.log);
+  }
+  render() {
+    return (
+      <Fragment>
+        {!this.state.loaded ? (
+          <p>loading...</p>
```
justafish: s/.../… :neckbeard:  🤓

src/components/05_pages/Permissions/Permissions.jsx
dawehner: Should we create a quick ticket to track this? For me ideally we write the data one day back to Drupal

src/components/05_pages/Permissions/Permissions.jsx
dawehner: There is a bit of a pane involved with CRE here. Environment variables aren't passed along unless they are prefixed with react. I had this problem when working on my config experiment, see https://github.com/jsdrupal/drupal-admin-ui/compare/config-experiment-2#diff-01454d5c4c96a618ac33dd690eb3c20bR128

dawehner: I hope we can workaround that though :)

src/components/05_pages/Permissions/Permissions.jsx
dawehner: I'm wondering whether there is a better thing to show.

src/components/05_pages/Permissions/Permissions.jsx
justafish: oh yeah, I forgot about that. I think prefixing with REACT_ is totally fine :+1:

src/components/05_pages/Permissions/Permissions.jsx
justafish: a party parrot, obviously

package.json
mattgrill: Eslint is included with CRA, but none of the other deps we need for airbnb or prettier plugins.

src/App.jsx
mattgrill: Yes?

.eslintrc.json
mattgrill: You can put JSX in .js files.

package.json
justafish: CRA uses eslint:recommended instead (see https://github.com/facebook/create-react-app/issues/3540#issuecomment-358792632), but I think we can avoid this when we bump to react-scripts 2.0 (if we do want airbnb here).
Can we split this eslint/prettier stuff into a separate PR so we can continue there and not hold things up here.

.eslintrc.json
dawehner: Should we stick with one way? At work we use JS, because well, that's what is in there :)

justafish: I'm fine with that, but I think all this stuff needs moving over to a separate issue to discuss how/where we deviate from CRA

src/components/05_pages/Permissions/Permissions.js
```diff
@@ -1,17 +1,98 @@
...
+        .reduce((acc, cur) => {
+          acc[cur.provider] = acc[cur.provider] || [];
+          acc[cur.provider].push(cur);
+          return acc;
+        }, {}),
+    );
+  render() {
+    return (
+      <Fragment>
+        {!this.state.loaded ? (
```
dawehner: Here is a generic question. Is there a reason why you do the condition inside the fragment and not outside of it?

src/components/05_pages/Permissions/Permissions.js
```diff
@@ -1,17 +1,98 @@
...
-const styles = {
-  title: css`
-    text-decoration: underline;
-  `,
+const Permissions = class Permissions extends Component {
+  state = {
+    loaded: false,
+  };
+  componentDidMount() {
+    Promise.all([
```
dawehner: Also a general question. Don't we need some handling for the case that the component is not longer active but the promise resolves? (ideally we would cancel the promise as part of ```componentDidUnmount``` or something like that.

src/components/05_pages/Permissions/Permissions.js
mattgrill: Since we're just returning the `<table />` here we probably don't even need the fragment at all.

src/components/05_pages/Permissions/Permissions.js
mattgrill: Yes, we could set the promise up in the constructor and just execute when the component mounts, and cancel it later.

dawehner: Let's keep that up as a todo, I guess?

mattgrill: I'll just get these wrapped up today. :)

src/components/05_pages/Permissions/Permissions.js
```diff
@@ -1,17 +1,112 @@
...
+        .then(([permissions, { data: roles }]) =>
+          this.setState({ permissions, roles, loaded: true }),
+        )
+        .catch(err => this.setState({ loaded: false, err })),
+    );
+  }
+  componentWillUnmount() {
+    this.cancelFetch();
+  }
+  fetchData = () =>
```
dawehner: I think we can drop this now :)

src/components/05_pages/Permissions/Permissions.js
```diff
@@ -1,17 +1,112 @@
...
+              ),
+            ].map(label => <td key={`column-${label}`}>{label}</td>)}
+          </tr>
+        </thead>
+        <tbody>
+          {this.groupPermissions(this.state.permissions).map(
+            ([permissionGroupName, permissions]) =>
+              permissions.length && (
+                <Fragment key={`fragment-${permissionGroupName}`}>
+                  <tr key={`permissionGroup-${permissionGroupName}`}>
```
dawehner: Do we need key on both levels?

src/components/05_pages/Permissions/Permissions.js
```diff
@@ -1,17 +1,112 @@
...
+                      }`}
+                    >
+                      <td>{permission.title}</td>
+                      {this.state.roles.map(({ attributes }) => (
+                        <td
+                          key={`role-${attributes.id}-permission-${
+                            permission.id
+                          }`}
+                        >
+                          {attributes.is_admin &&
```
dawehner: Nice level of detail!

src/components/05_pages/Permissions/Permissions.js
mattgrill: Yep. Which is strange, but each element within the loop needs a key.

src/components/05_pages/Permissions/Permissions.js
mattgrill: 👍

src/components/UI/table.js
```diff
@@ -0,0 +1,47 @@
...
+};
+TABLE.propTypes = {
+  children: oneOfType([arrayOf(node), node]).isRequired,
+  zebra: bool,
+};
+TABLE.defaultProps = {
+  zebra: false,
+};
+
+const TBODY = ({ children, ...props }) => <tbody {...props}>{children}</tbody>;
```
dawehner: Oh default HTML elements don't offer proptypes?

src/components/05_pages/Permissions/Permissions.js
```diff
@@ -1,17 +1,112 @@
...
+  fetchData = () =>
+    Promise.all([
+      fetch(
+        `${process.env.REACT_APP_DRUPAL_BASE_URL}/permissions?_format=json`,
+      ).then(res => res.json()),
+      fetch(
+        `${process.env.REACT_APP_DRUPAL_BASE_URL}/jsonapi/user_role/user_role`,
+      ).then(res => res.json()),
+    ]);
+  groupPermissions = permissions =>
```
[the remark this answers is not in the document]
mattgrill: One thing we will need to figure out, is how to handle polyfills. Since this uses `entries` we will want to figure that out.

src/components/UI/table.js
mattgrill: So I wanted to provide these wrappers so if we want to provide styling/functionality for basic DOM elements, we can do it all in one place and those changes will be reflected everywhere.

src/components/05_pages/Permissions/Permissions.js
```diff
@@ -1,17 +1,94 @@
...
+                          checked={attributes.permissions.includes(
+                            permission.id,
+                          )}
+                        />
+                      ),
+                  ),
+                ],
+              })),
+            ])
+            .reduce((a, b) => a.concat(b), [])}
```
dawehner: Note: You can use ```[].concat(...a)``` instead of the reduce.

src/components/Helpers/Loading.js
```diff
@@ -0,0 +1,58 @@
...
+    40% {
+      -webkit-transform: scale(1);
+      transform: scale(1);
+    }
+  }
+`;
+
+const Loading = () => (
+  <div className={spinner}>
+    <div className="bounce1" />
```
justafish: I've had some really Bad Times(tm) with using class names this way when the project starts to grow. The best way we've found is to have an object called styles and then do `className={styles.spinner}`, `className={styles.bounce1}` etc

mattgrill: Yep, this totally makes sense.

src/components/05_pages/Permissions/Permissions.js
```diff
@@ -108,7 +100,7 @@ const Permissions = class Permissions extends Component {
                 ) : (
                   <input
                     type="checkbox"
-                    onChange={e =>
+                    onChange={() =>
```
dawehner: I'm wondering: Is there a way to avoid creating new functions for each rendered checkbox?

src/components/Helpers/Loading.js
```diff
@@ -5,28 +5,32 @@ const styles = {
...
-    }
-  `,
-  right: css`
-    ::before {
-      content: '➡️';
-    }
-  `,
   peace: css`
-    ::before {
-      content: '✌️';
```
mherchel: ✌️

src/components/Helpers/Loading.js
```diff
@@ -0,0 +1,37 @@
...
+    display: inline-block;
+    vertical-align: middle;
+    animation-direction: alternate;
+    animation-iteration-count: infinite;
+    animation-duration: 0.5s;
+    animation-timing-function: cubic-bezier(0, 0, 1, 1);
+    transform-origin: bottom;
+    font-size: 50px;
+    animation-name: rotate;
+    @keyframes rotate {
```
mherchel: Does this actually work? I've done keyframes in emotion according to https://github.com/emotion-js/emotion/blob/master/docs/keyframes.md

```js
const rotate = keyframes`
  //...
`
```

mattgrill: Yes, it does work.

loading

mherchel: ✌️

src/components/05_pages/Permissions/Permissions.js
mattgrill: @dawehner Let me investigate that.

src/components/UI/table.js
```diff
@@ -0,0 +1,81 @@
...
+);
+THEAD.propTypes = {
+  data: arrayOf(string).isRequired,
+};
+
+const TBODY = ({ rows }) => (
+  <tbody>
+    {rows.map(({ colspan, tds, key }) => (
+      <TR key={key}>
+        {generateIDs(tds).map(({ id, value }) => (
```
justafish: What's the advantage of doing this over using the array index?

src/components/UI/index.js
```diff
@@ -0,0 +1,9 @@
+import { TR, TD, TABLE, TBODY, THEAD } from './table';
+
+export {
```
justafish: Why do we do this instead of directly importing from a table file?

mattgrill: I did this so if we ever have any other components, they are all importable from a single file.

src/components/UI/table.js
mattgrill: React gets sad if you use the index as part of the key.

src/components/Helpers/Loading.js
mrjmd: This PR is amazing

src/components/UI/index.js
dawehner: Personally I think we should come back at some point and have a look at the bigger picture to understand how we can make the entire app more readable and not just single bits and pieces.

justafish: I had a look at some other libraries and the lodash method seems to be popular where you pull in the thing you need like `import { td, tr } from 'ui/table'`, but yeah we can address this in a follow up, doesn't need to block this PR

src/components/05_pages/Permissions/Permissions.js
```diff
@@ -1,17 +1,156 @@
...
+      Promise.all([
+        fetch(
+          `${
+            process.env.REACT_APP_DRUPAL_BASE_URL
+          }/admin-api/permissions?_format=json`,
+        ).then(res => res.json()),
+        fetch(
+          `${
+            process.env.REACT_APP_DRUPAL_BASE_URL
+          }/jsonapi/user_role/user_role`,
```
e0ipso: You probably want to add the `Accept: application/vnd.api+json` header. This will enable 404/403 errors in JSON format (if any). Otherwise they'll come back as HTML errors.

src/components/05_pages/Permissions/Permissions.js
```diff
@@ -1,17 +1,156 @@
...
+    Object.entries(
+      permissions.reduce((acc, cur) => {
+        acc[cur.provider] = acc[cur.provider] || [];
+        acc[cur.provider].push(cur);
+        return acc;
+      }, {}),
+    );
+
+  createTableRows = (groupedPermissions, roles) =>
+    [].concat(
```
e0ipso: Why `[].concat()` instead of array spreads? It seems you prefer spreads elsewhere.

e0ipso:
Pretty awesome. I almost followed most of it.

src/components/05_pages/Permissions/Permissions.js
e0ipso: For clarity, I'm not after any change here, I'm genuinely curious,

mattgrill: @e0ipso It's a bit subtle, but this is the easiest way to flatten the two-dimensional array that's created from constructing the table rows.

src/components/05_pages/Permissions/Permissions.js
mattgrill: 👍

src/components/UI/index.js
justafish: @mattgrill cool cool, let's do away with index.js then. Not sure what we should have the exports as exactly? My preference would be:`TD`, `TR`, `Table`, `TBody`, and `THead`

mattgrill:
Based on the conversation, perhaps we can merge this today?

src/components/Helpers/Loading.js
mattgrill: I think we resolved this?

mattgrill:
permissions

mattgrill:
A small highlight of the current permissions page work.

PR #10, opened 2018-03-08
Prettier

mattgrill:
- prettier
- run prettier


.prettierrc.json
```diff
@@ -0,0 +1,6 @@
+{
+  "printWidth": 80,
+  "semi": true,
+  "singleQuote": true,
+  "trailingComma": "all"
```
dawehner: I love trailing comma all!

.eslintrc.json
```diff
@@ -0,0 +1,17 @@
+{
+  "extends": [
+    "eslint-config-airbnb",
+    "plugin:prettier/recommended"
```
dawehner: I tried to compare what eslint-plugin-prettier does: https://github.com/prettier/eslint-plugin-prettier#recommended-configuration vs. what https://github.com/facebook/create-react-app/blob/next/packages/eslint-config-react-app/index.js does. Given we already try to go down the airbnb ruleset in core, I think using the same here makes 100% sense.

package.json
```diff
@@ -9,9 +9,19 @@
...
+    "eslint-config-prettier": "^2.9.0",
+    "eslint-plugin-import": "^2.7.0",
+    "eslint-plugin-jsx-a11y": "^6.0.2",
+    "eslint-plugin-prettier": "^2.6.0",
+    "eslint-plugin-react": "^7.4.0",
+    "prettier": "^1.11.1",
     "react-scripts": "1.1.1"
   },
   "scripts": {
+    "prettier": "./node_modules/.bin/prettier --write 'src/**/*.jsx' 'src/**/*.js'",
```
dawehner: Can't you just use "prettier"? It'll pick up things in node_modules/.bin/prettier automatically

mattgrill: Yeah, you can. I just always try and be specific.

.eslintrc.json
```diff
@@ -0,0 +1,17 @@
...
+  "parser": "babel-eslint",
+  "root": true,
+  "env": {
+    "browser": true,
+    "es6": true
+  },
+  "rules": {
+    "no-param-reassign": [0],
+    "no-underscore-dangle": [0],
+    "react/jsx-filename-extension": [1, { "extensions": [".js", ".jsx"] }]
```
justafish: Let's ditch jsx completely

mattgrill:
@justafish I just looked into this, if you add an .eslintrc.json it picks it up.

.eslintrc.json
mattgrill: We can, however there are some tests that are `.js` but they contain JSX. I don't care either way which is why I added this.

.eslintrc.json
```diff
@@ -0,0 +1,15 @@
+{
+  "extends": [
+    "eslint-config-airbnb",
+    "plugin:prettier/recommended"
+  ],
+  "parser": "babel-eslint",
+  "root": true,
+  "env": {
+    "browser": true,
+    "es6": true
```
justafish: + jest, and serviceworker

justafish:
Just the linting errors themselves that need fixing now :+1:

justafish:
So one thing I don't understand about adding `eslint-config-prettier` is it says that it "Turns off all rules that are unnecessary or might conflict with Prettier." - so what is enforcing the rules that it's turned off? Running prettier is something that each person needs to have done manually (or setup to do automatically on save, but still). If that is the case, then I would have thought our prettier config would be setup to help you follow the standards we've adopted (airbnb in this case), and so I'm not sure what this plugin actually does?

justafish:
Ok I read https://prettier.io/docs/en/eslint.html and I get it now :+1:

Issue #18, opened 2018-03-14
Should API responses from support module include HTML in label, descriptions, etc


tedbow:
Working on https://github.com/jsdrupal/drupal-admin-ui-support/pull/1 @dawehner @mattgrill and I were dicussing whether the permissions titles and descriptions should have the HTML stripped out on the server-side.

One thing to keep in mind is that eventually we would want core itself to provide these endpoints. So should make them now as they would likely be added to core?

dawehner:
For a generic endpoint I think having both would be nice. I totally understand that some usecases will not use the HTML, but for others it could be nice UI wise.

lauriii:
I think for now it is fine if we have the HTML in the API responses. In many non-web contexts, it is still possible to render HTML. Also, if someone would need the response contents without HTML, they could always filter it out.

Issue #4, opened 2018-03-04
Write a composer script that will insert the admin theme into Drupal


justafish:
The admin theme is available on packagist now: https://packagist.org/packages/justafish/drupal-admin-ui

The install command will need to create a symlink (or create a copy) from `vendor/justafish/drupal-admin-ui` to `vfancy`

dawehner:
Research so far:

* Packages can't have install hooks (probably security reasons)
* You can create custom installers though, see https://getcomposer.org/doc/articles/custom-installers.md#creating-an-installer
* Given that we would probably need a repo containing the custom installer command.
* You would need to add this custom installer composer plugin into the root dir of Drupal and then install this package.

justafish:
There's a big discussion here: https://github.com/composer/composer/issues/1193

@alexpott, @dawehner and I worked on this and came up with https://www.drupal.org/project/drupal/issues/2949832

ryanaslett:
If Im understanding the patch in https://www.drupal.org/project/drupal/issues/2949832 correctly, its moving the file out of vendor and into the web folder so that it can be found at 'https://sitename/vfancy' ?

There exists a mechanism in composer to do this: 
If we define a new `type` key, for example, 'drupal-js-ui', and define a folder they can go into, then we can submit a PR to composer/Installers (https://github.com/composer/installers/blob/master/src/Composer/Installers/DrupalInstaller.php)
`'drupal-js-ui'   => 'js-admin-ui/{$name}/',`

When you `composer require jsdrupal/admin-ui` it will install it to the proper location put it where it belongs.

The only caveat there is that it wouldn't put it directly in the web folder so the url would end up being `'https://sitename/js-admin-ui/vfancy'`, but, if we may have more than one react apps built on top of drupal's api someday, that seems like we might want that anyhow.

mglaman:
What if the app assumed they followed composer flow where vendor is outside of docroot? So it'd be

* docroot / web
* vendor
* jsdrupal/vfancy

Two docroot, then?

ryanaslett:
If they're modifying where doc root is, then they're also modifying the composer installers locations, much like the drupal-project already does: https://github.com/drupal-composer/drupal-project/blob/8.x/composer.json#L59

alexpott built this: https://github.com/jsdrupal/test-project  using the oomphinc/composer-installers-extender which defines where drupal-admin-ui gets installed to.

justafish:
[...]

PR #20, opened 2018-03-14
Move the support module into the main repo

dawehner:
Note: 
There is https://github.com/jsdrupal/drupal-admin-ui/issues/18 and
https://github.com/jsdrupal/drupal-admin-ui/issues/4 to simplify the setup in the future.

### DIFF
--- a/admin_ui_support/src/Plugin/rest/resource/PermissionsCollectionResource.php
+++ b/admin_ui_support/src/Plugin/rest/resource/PermissionsCollectionResource.php
@@ -105,9 +105,6 @@ class PermissionsCollectionResource extends ResourceBase {
       $permissions = $this->permissionHandler->getPermissions();
       foreach ($permissions as $id => $permission) {
         $permissions[$id]['id'] = $id;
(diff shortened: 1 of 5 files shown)


admin_ui_support/src/Plugin/rest/resource/PermissionsCollectionResource.php
```diff
@@ -0,0 +1,137 @@
...
+  public function get() {
+    $context = new RenderContext();
+    
+    $permissions = $this->renderer->executeInRenderContext($context, function () {
+      $permissions = $this->permissionHandler->getPermissions();
+      foreach ($permissions as $id => $permission) {
+        $permissions[$id]['id'] = $id;
+        // @todo Should HTML be exposed?  https://github.com/jsdrupal/drupal-admin-ui/issues/18
+        $permissions[$id]['title'] = PlainTextOutput::renderFromHtml($permission['title']);
+        $permissions[$id]['description'] = PlainTextOutput::renderFromHtml($permission['description']);
```
mattgrill: I think this is what needs to be removed? @tedbow

PR #22, opened 2018-03-19
Make tables look less awful and easier to use.

mattgrill:
- Make tables less miserable to look at.
- Slight file system organizational change.


mattgrill:
react_app

PR #24, opened 2018-03-20
Try to improve the dev instructions (the Drupal installation bit isn't there yet)

dawehner:


README.md
```diff
@@ -15,11 +15,21 @@ cd ../drupal
 ln -s ../drupal-admin-ui/build vfancy
 
 ## Link the support module to Drupal
-cd ../drupal
-ln -s ../drupal-admin-ui/admin_ui_support modules
+cd ../drupal/modules
+ln -s ../../drupal-admin-ui/admin_ui_support .
+
+## Install the jsonapi module
```
tedbow: We should say "Download" here because we "Install" bellow.

README.md
```diff
@@ -15,11 +15,21 @@ cd ../drupal
...
+curl -O https://ftp.drupal.org/files/projects/jsonapi-8.x-1.13.tar.gz
+tar zxf jsonapi-8.x-1.13.tar.gz
 
-# Install Drupal and start webserver
+# Install Drupal
+@todo Use the dev-site command
 composer install
+
+## Install the needed modules
+drush en jsonapi admin_ui_support -y
```
tedbow: In https://github.com/jsdrupal/drupal-admin-ui/pull/29 I added jsonapi as a dependency so that would here we would only need `admin_ui_support`

README.md
dawehner: +1 Do you want to fix it?

README.md
dawehner: Works for me! Thank you

dawehner:
Let's merge it :)

PR #23, opened 2018-03-19
WIP: Expose the menu tree for the admin toolbar

dawehner:
Example output:

```
{
    "49990 Content system.admin_content": {
        "subtree": {
            "50000 Comments comment.admin": {
                "subtree": [],
                "hasChildren": false,
                "inActiveTrail": false,
                "options": {
                    "attributes": {
                        "id": "toolbar-link-comment-admin",
                        "class": [
                            "toolbar-icon",
                            "toolbar-icon-comment-admin"
                        ],
                        "title": "List and edit site comments and the comment approval queue."
                    }
                },
                "link": {
                    "weight": "0",
                    "title": "Comments",
                    "description": "List and edit site comments and the comment approval queue.",
                    "menuName": "admin",
                    "url": "/d8/admin/content/comment"
                }
            }
        },
        "hasChildren": true,
        "inActiveTrail": false,
        "options": {
            "attributes": {
                "id": "toolbar-link-system-admin_content",
                "class": [
                    "toolbar-icon",
                    "toolbar-icon-system-admin-content"
                ],
                "title": "Find and manage content."
            }
        },
        "link": {
            "weight": "-10",
            "title": "Content",
            "description": "Find and manage content.",
            "menuName": "admin",
            "url": "/d8/admin/content"
        }
    },
    "49992 Structure system.admin_structure": {
        "subtree": {
            "50000 Block layout block.admin_display": {
                "subtree": [],
                "hasChildren": false,
                "inActiveTrail": false,
                "options": {
                    "attributes": {
                        "id": "toolbar-link-block-admin_display",
                        "class": [
                            "toolbar-icon",
                            "toolbar-icon-block-admin-display"
                        ],
                        "title": "Configure what block content appears in your site's sidebars and other regions."
                    }
                },
                "link": {
                    "weight": "0",
                    "title": "Block layout",
                    "description": "Configure what block content appears in your site's sidebars and other regions.",
                    "menuName": "admin",
                    "url": "/d8/admin/structure/block"
                }
            },
            "50000 Comment types entity.comment_type.collection": {
                "subtree": [],
                "hasChildren": false,
                "inActiveTrail": false,
                "options": {
                    "attributes": {
                        "id": "toolbar-link-entity-comment_type-collection",
                        "class": [
                            "toolbar-icon",
                            "toolbar-icon-entity-comment-type-collection"
                        ],
                        "title": "Manage form and displays settings of comments."
                    }
                },
                "link": {
                    "weight": "0",
                    "title": "Comment types",
[...]

PR #25, opened 2018-03-20
Add nav menu

justafish:
Adds a navigation menu (with hardcoded links for now) and some (temporary) bad CSS so it's usable

ezgif-3-1c6e5fee5d

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,8 +1,9 @@
 import React, { Component } from 'react';
-import { BrowserRouter as Router, Route, Link } from 'react-router-dom';
+import { BrowserRouter as Router, Route, Switch } from 'react-router-dom';
 
 import routes from './routes';
 
+import Default from './components/06_wrappers/Default/Default';
 import Home from './components/05_pages/Home/Home';
 import NoMatch from './NoMatch';
 
@@ -16,30 +17,15 @@ class App extends Component {
   render() {
     return (
       <Router>
-        <div>
-          <ul>
-            <li>
-              <Link to="/">Home</Link>
-            </li>
-            <li>
-              <Link to="/admin/people/permissions">Permissions</Link>
-            </li>
-            <li>
-              <Link to="/admin/appearance">Appearance</Link>
-            </li>
-            <li>
-              <Link to="/node/add">Content</Link>
-            </li>
-          </ul>
-
-          <hr />
-
-          <Route exact path="/" component={Home} />
-          {Object.keys(routes).map(route => (
-            <Route path={route} component={routes[route]} key={route} />
-          ))}
-          <Route component={NoMatch} />
-        </div>
+        <Default>
+          <Switch>
+            <Route exact path="/" component={Home} />
+            {Object.keys(routes).map(route => (
+              <Route path={route} component={routes[route]} key={route} />
+            ))}
+            <Route component={NoMatch} />
+          </Switch>
+        </Default>
       </Router>
     );
   }
--- a/src/components/05_pages/Permissions/Permissions.js
+++ b/src/components/05_pages/Permissions/Permissions.js
@@ -19,12 +19,20 @@ const Permissions = class Permissions extends Component {
           `${
             process.env.REACT_APP_DRUPAL_BASE_URL
           }/admin-api/permissions?_format=json`,
+          {
+            credentials: 'include',
+          },
         ).then(res => res.json()),
         fetch(
           `${
             process.env.REACT_APP_DRUPAL_BASE_URL
           }/jsonapi/user_role/user_role`,
-          { headers: { Accept: 'application/vnd.api+json' } },
+          {
+            headers: {
+              Accept: 'application/vnd.api+json',
+            },
+            credentials: 'include',
+          },
         ).then(res => res.json()),
       ])
         .then(([permissions, { data: roles }]) =>
--- a/src/components/06_wrappers/Default/Default.js
+++ b/src/components/06_wrappers/Default/Default.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import PropTypes from 'prop-types';
+import { node } from 'prop-types';
 import { css } from 'emotion';
 import { scaleRotate as Menu } from 'react-burger-menu';
 import { Link } from 'react-router-dom';
@@ -72,7 +72,7 @@ const Default = props => (
 );
 
 Default.propTypes = {
-  children: PropTypes.node.isRequired,
+  children: node.isRequired,
 };
 
 styles = {
--- a/src/components/06_wrappers/Default/Default.js
+++ b/src/components/06_wrappers/Default/Default.js
@@ -1,0 +1,129 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { css } from 'emotion';
+import { scaleRotate as Menu } from 'react-burger-menu';
+import { Link } from 'react-router-dom';
+
+let styles;
+
+const menuLinks = {
+  home: [{ path: '/', text: '🏠 Home' }],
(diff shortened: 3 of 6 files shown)


src/components/06_wrappers/Default/Default.js
```diff
@@ -0,0 +1,130 @@
...
+import { css } from 'emotion';
+import { scaleRotate as Menu } from 'react-burger-menu';
+import { Link } from 'react-router-dom';
+
+let styles;
+
+const menuLinks = {
+  home: [{ path: '/', text: '🏠 Home' }],
+  permissions: [{ path: '/admin/people/permissions', text: '🔐 Permissions' }],
+  content: [
```
dawehner: I'm a bit confused that you are not using the same structure as the PR in https://github.com/jsdrupal/drupal-admin-ui/pull/23

src/components/05_pages/Permissions/Permissions.js
```diff
@@ -19,12 +19,20 @@ const Permissions = class Permissions extends Component {
           `${
             process.env.REACT_APP_DRUPAL_BASE_URL
           }/admin-api/permissions?_format=json`,
+          {
```
dawehner: Do you mind moving this into its own PR :)

src/components/06_wrappers/Default/Default.js
justafish: I hadn't seen that yet, this is only supposed to be temporary as noted

dawehner: Sure, it'd be nice to come up with a usecase based structure on the other PR though

dawehner:
It'd be nice to not be too playfull here. Otherwise people start to nitpick on that.

src/components/06_wrappers/Default/Default.js
justafish: I was going to wait until the other PR is committed and then create a new one to use the new endpoint

src/components/05_pages/Permissions/Permissions.js
justafish: Can we leave it for now? I was going to work on a follow up PR that puts all this API stuff into a separate wrapper file as there's a lot of repetition here

justafish:
@dawehner styling is totally temporary, just wanted to get something in place so the menu is usable

dawehner:
Well, just remove the crazy effect :)

justafish:
nooo @dawehner that's my favourite part! 😂 I chose that one because it looks nice, shows off cool things we can do with components from npm, and doesn't require the page to be responsive (unlike the one which squashes stuff in, for example).
Also, just want to re-iterate that all the styling here is temporary! If this inspires people to come and nitpick on it and make something better than that's a +1 as far as I'm concerned 😁

src/components/06_wrappers/Default/Default.js
dawehner: Fair :) I also just asked @tedbow to maybe think about a better structure than what is right now in the other PR, so its good to not be blocked on this right now.

dawehner:
Let's just use the crazy effect right now and then get the UX team actually involved. I think this effect shows that its potentially not the end of what we want :)

src/components/05_pages/Permissions/Permissions.js
dawehner: I'd suggest to not abstract till we have another usecase.

dawehner: I would have suggest @mattgrill to work on it, given he has actually written a library doing exactly that.

dawehner: To clarify, I wanted to express: @mattgrill would be a good candidate when you don't have time to work on it.

src/components/06_wrappers/Default/Default.js
mattgrill: @justafish @dawehner I mentioned that we should just merge the menu pr now so we can continue to quickly iterate on this.

src/components/05_pages/Permissions/Permissions.js
mattgrill: @justafish @dawehner I don't mind working on this less-than-glamorous part. I was already planning on doing something like this when I do the changes to update Drupal with the permissions changes.

mattgrill: Additionally, I think this is fine to add here. I'll probably just copy all this out into it's own helper anyways.

mattgrill:
One of the things we've tried in the past is to add a reduced animation mode. I think that having something like that, that you can configure would be nice. I'm not sure it's a blocker for this though. Maybe an issue/todo we can add.

src/App.js
```diff
@@ -16,30 +17,15 @@ class App extends Component {
...
-          <hr />
-
-          <Route exact path="/" component={Home} />
-          {Object.keys(routes).map(route => (
-            <Route path={route} component={routes[route]} key={route} />
-          ))}
-          <Route component={NoMatch} />
-        </div>
+        <Default>
+          <Switch>
```
mattgrill: This is great!

src/components/06_wrappers/Default/Default.js
mattgrill: It might be worth it to add some sort of normalizer to the returned JSON from #23 to get to this point?

src/components/06_wrappers/Default/Default.js
```diff
@@ -0,0 +1,130 @@
...
+    </Menu>
+
+    <main className={styles.main} id={styles.main}>
+      {props.children}
+    </main>
+  </div>
+);
+
+Default.propTypes = {
+  children: PropTypes.node.isRequired,
```
mattgrill: In `Permissions` I dropped the `PropTypes` bit and just used the exact element type. It comes out looking a bit cleaner.

src/components/06_wrappers/Default/Default.js
```diff
@@ -0,0 +1,130 @@
...
+};
+
+styles = {
```
mattgrill: Do we want to maybe but these in a different file?

justafish: I like having them together - like JS and HTML being in the same file 😁

src/components/06_wrappers/Default/Default.js
justafish: oh, I'm surprised the linter let you do that! (given https://github.com/facebook/prop-types/issues/150)

mattgrill: Yep. Works just fine.

src/components/06_wrappers/Default/Default.js
mattgrill: I can only see this getting more unmanageable as each of these files gets more complicated.

src/components/06_wrappers/Default/Default.js
justafish: :+1:

src/components/06_wrappers/Default/Default.js
justafish: @mattgrill we haven't had this issue when using it in the wild - generally they end up being split into smaller components if that's becoming the case. It's also much easier to work with all in the same place when you start using props inside your styles, styled components etc - much like not having a separate HTML and JS file anymore.

src/components/05_pages/Permissions/Permissions.js
justafish: :+1:

src/components/06_wrappers/Default/Default.js
justafish: :+1:

justafish:
@mattgrill PR updated with destructured propTypes

PR #29, opened 2018-03-20
Override entity resource handling in jsonapi for config entities

tedbow:
We will need to be able to POST/PATCH config entities in JSON API. 

Currently this not explicitly supported. PATCH seemed to work but POST fails because of field access checks.

@see https://www.drupal.org/project/jsonapi/issues/2887313

This PR introduces a ConfigEntityResource that will allow POST/PATCH on config entities. 

This currently does no validation.


tedbow:
We could add basic validation like:
`/** @var \Drupal\Core\Config\TypedConfigManagerInterface $type_config_manager */
+    $type_config_manager = \Drupal::service('config.typed');
+    $typed_config = $type_config_manager->createFromNameAndData($entity->getConfigDependencyName(), $entity->toArray());
+    $violations = $typed_config->validate();`

admin_ui_support/src/Controller/ConfigEntityResource.php
```diff
@@ -0,0 +1,55 @@
...
+
+  /**
+   * {@inheritdoc}
+   *
+   * Overridden to remove validation which isn't supported by config entitie.
+   *
+   * @see https://www.drupal.org/project/drupal/issues/2300677
+   */
+  public function createIndividual(EntityInterface $entity, Request $request) {
+    $entity_access = $entity->access('create', NULL, TRUE);
```
dawehner: I'm curious, what happens on the update case? Does this just work?

tedbow: Yes works because when patching an entity `\Drupal\jsonapi\Controller\EntityResource::updateEntityField` first checks if the field is a `ContentEntityInterface` before checking field access whereas `\Drupal\jsonapi\Controller\EntityResource::createIndividual` just checks all fields for access.

This probably an oversight so since from my understanding it is not the intention of Jsonapi to support POST/PATCH on config entities. So we would probably have to expand `\Drupal\admin_ui_support\Controller\ConfigEntityResource` at some point to override `::createIndividual()`. 

Also we may have to alter the routes to produce because I am not sure why jsonapi even supports POST/PATCH methods for config entities.

dawehner: Fair, but isn't there also ```\Drupal\jsonapi\Controller\EntityResource::patchIndividual``` which should be the code path triggered here?

tedbow: Yes but `\Drupal\jsonapi\Controller\EntityResource::patchIndividual` calls which is where the actual field access check happens.

PR #48, opened 2018-03-25
Redirect to the app

dawehner:
Test instructions:

**Blocked by https://github.com/jsdrupal/drupal-admin-ui/pull/55 right now**

* Enable the module
* Ensure your-host/vfancy works in general
* Goto your-host/admin/people/permissions
* Ensure that the vfancy UI appears with the permission form preselected

### DIFF
--- a/admin_ui_support/admin_ui_support.routing.yml
+++ b/admin_ui_support/admin_ui_support.routing.yml
@@ -1,0 +1,27 @@
+admin_ui_support.user.permissions:
+  path: '/vfancy/admin/people/permissions'
+  defaults:
+    _controller: '\Drupal\admin_ui_support\Controller\EmptyController::noop'
+  options:
+    _admin_related_route: 'user.admin_permissions'
+  requirements:
+    _permission: 'administer permissions'
+
+admin_ui_support.entity.user_role.edit_permissions_form:
+  path: '/vfancy/admin/people/permissions/{user_role}'
+  defaults:
+    _controller: '\Drupal\admin_ui_support\Controller\EmptyController::noop'
+  options:
+    _admin_related_route: 'entity.user_role.edit_permissions_form'
+  requirements:
+    _entity_access: 'user_role.update'
+
+admin_ui_support.entity.user_role.collection:
+  path: '/vfancy/admin/people/roles'
+  defaults:
+    _controller: '\Drupal\admin_ui_support\Controller\EmptyController::noop'
+  options:
+    _admin_related_route: 'entity.user_role.collection'
+  requirements:
+    _permission: 'administer permissions'
+
--- a/admin_ui_support/src/Routing/RouteSubscriber.php
+++ b/admin_ui_support/src/Routing/RouteSubscriber.php
@@ -1,0 +1,54 @@
+<?php
+
+namespace Drupal\admin_ui_support\Routing;
+
+use Drupal\Core\Routing\RouteSubscriberBase;
+use Drupal\Core\Routing\RoutingEvents;
+use Symfony\Component\Routing\RouteCollection;
+
+/**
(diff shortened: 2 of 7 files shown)


src/App.js
```diff
@@ -10,13 +11,17 @@ import NoMatch from './NoMatch';
 import normalize from './styles/normalize'; // eslint-disable-line no-unused-vars
 import base from './styles/base'; // eslint-disable-line no-unused-vars
 
+const history = createBrowserHistory();
+
 class App extends Component {
   componentDidMount() {
-    window.history.replaceState(null, null, '/');
+    // Allow Drupal redirects to determine the initial path.
+    const search = history.location.search.replace('?q=', '');
```
dawehner: This bit ensures that the initial route is the one requested from Drupal. Also: I removed window.history. We should rather use the react router history

tedbow:
I don't think i can review this before https://github.com/jsdrupal/drupal-admin-ui/pull/55 is merged because the app is not working for me.

dawehner:
Agreed!

dawehner:
Rerolled it now

dawehner:
@tedbow Do you have questions about this PR?

src/App.js
```diff
@@ -40,7 +40,9 @@ const withDefault = component => () => (
 
 class App extends Component {
   componentDidMount() {
-    // history.replace('/');
+    // Allow Drupal redirects to determine the initial path.
+    const search = history.location.search.replace('?q=', '');
```
lauriii: Do we really want to change the actual location (as in make another redirect)?

I also tried to navigate to /vfancy/admin/people/permissions path and it doesn't seem to be working.
screen shot 2018-03-30 at 23 40 40

admin_ui_support/src/Routing/RouteSubscriber.php
```diff
@@ -0,0 +1,54 @@
...
+ * Listens to the dynamic route events.
+ */
+class RouteSubscriber extends RouteSubscriberBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function alterRoutes(RouteCollection $collection) {
+    $take_over_routes = [];
+    $call_back_routes = [];
```
lauriii: What is this?

src/App.js
dawehner: > I also tried to navigate to /vfancy/admin/people/permissions path and it doesn't seem to be working.

dawehner: I'm not sure what to do. We could actually remove the original routes from the router. This is really just an implementation detail, which we might come up with a better one in the future.

admin_ui_support/src/Routing/RouteSubscriber.php
dawehner: I simplified the code quite a bit

admin_ui_support/admin_ui_support.routing.yml
```diff
@@ -0,0 +1,27 @@
+admin_ui_support.user.permissions:
+  path: '/vfancy/admin/people/permissions'
+  defaults:
+    _controller: '\Drupal\admin_ui_support\Controller\EmptyController::noop'
+  options:
+    _admin_related_route: 'user.admin_permissions'
```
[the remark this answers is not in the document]
lauriii: I was having same thoughts but I thought it might be overkill to go with any of the other approaches for now.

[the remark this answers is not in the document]
dawehner: Well, please tell them :)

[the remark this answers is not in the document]
dawehner: One day I think this could also include a pointer to a JS component which should be rendered for this page.

PR #50, opened 2018-03-25
Add REST support for simple config

tedbow:
Convert patch from https://www.drupal.org/project/drupal/issues/2947759#comment-12508934 to create REST resources for simple config in support module


tedbow:
This still needs some work and I think we should add simple tests at least

tedbow:
Looking at RESTUI screen after this PR
screenshot 2018-03-25 12 35 33

Should our deriver actually create different uri's based on schema name like `\Drupal\rest\Plugin\Deriver\EntityDeriver` does for entity type? 

Otherwise we have a bunch of resources with the uri

dawehner:
@tedbow Really good point. I think generating a different route without parameters is the way to go for simple configuration. Nice

admin_ui_support/src/Plugin/rest/resource/SimpleConfigResource.php
```diff
@@ -139,6 +151,12 @@ public function availableMethods() {
    * {@inheritdoc}
    */
   public function calculateDependencies() {
+    // @todo Is there a better way to determine the module that provides the
+    // config?
+    $module_name = explode('.', $this->getPluginDefinition()['config_name'])[0];
+    if ($this->moduleHandler->moduleExists($module_name)) {
```
dawehner: Don't we want to add the dependencies all the time, no matter whether the module is installed or not?

admin_ui_support/src/Plugin/rest/resource/SimpleConfigResource.php
```diff
@@ -112,6 +113,22 @@ public function get() {
...
+  /**
+   * Responds to get request.
+   *
+   * @param array $data
+   *   The unserialized data.
+   *
+   * @return \Drupal\rest\ModifiedResourceResponse
+   *   The modified resource response.
+   */
+  public function patch(array $data) {
```
dawehner: Nice!

admin_ui_support/src/Plugin/Deriver/SimpleConfigDeriver.php
```diff
@@ -0,0 +1,120 @@
...
+
+    // Get all simple config and create plugin definition for them.
+    $config_schemas = $this->typedConfigManager->getDefinitions();
+    $simple_config_schemas = array_filter($config_schemas, function ($config_schema) use ($config_schemas) {
+      return $this->isConfigSchema($config_schema, $config_schemas);
+    });
+
+    foreach ($simple_config_schemas as $name => $simple_config_schema) {
+      // @todo Should we expose config with names that end in .*
+      if (substr($name, -2, 2) === '.*') {
```
larowlan: is the third arg needed here? kind of implied

admin_ui_support/src/Plugin/rest/resource/SimpleConfigResource.php
```diff
@@ -124,11 +127,31 @@ public function get() {
...
+  /**
+   * Responds to POST requests.
+   *
+   * @param array $data
+   *   The unserialized data.
+   *
+   * @return \Drupal\rest\ModifiedResourceResponse
+   *   The modified resource response.
+   */
+  public function post(array $data) {
```
dawehner: Does POST really make sense for simple configuration?

tedbow: Ok will remove

admin_ui_support/src/Plugin/Deriver/SimpleConfigDeriver.php
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    // Get all simple config and create plugin definition for them.
+    $config_schemas = $this->typedConfigManager->getDefinitions();
+    $simple_config_schemas = array_filter($config_schemas, function ($config_schema) use ($config_schemas) {
+      return $this->isConfigSchema($config_schema, $config_schemas);
+    });
+
+    foreach ($simple_config_schemas as $name => $simple_config_schema) {
+      // @todo Should we expose config with names that end in .*
```
dawehner: Well, I think this would no longer be simple configuration. Note: The code above already filters down to configuration.

dawehner:
I diid some additional minor changes. I'll just merge it for now to use it for generating the system site form

Issue #47, opened 2018-03-25
Add an API wrapper


justafish:
Consolidate the fetch calls we have into a central place

dawehner:
We have done that

PR #51, opened 2018-03-26
#47 Add redux saga wrappers around the API for fetching menu links

justafish:
_Now this is a story all about how
My life got flipped-turned upside down
And I'd like to take a minute
Just sit right there
I'll tell you how I found out that make-cancelable doesn't work_

tl;dr if you read further down [the thread that's linked from make-cancelable](https://github.com/facebook/react/issues/5465), there's an explanation for why it doesn't work.

So, I've added redux-saga because:
- It handles this really well
- We're going to need global state anyway
- It's [v](https://github.com/redux-saga/redux-saga/stargazers) [popular](https://www.npmjs.com/package/redux-saga)
- It's [v testable](https://redux-saga.js.org/docs/introduction/BeginnerTutorial.html#making-our-code-testable)

I also added a nice loading bar

This only partially completes #47, I just ported over the menu code and the Default template so this didn't become un-reviewable.


src/utils/api/api.js
```diff
@@ -0,0 +1,19 @@
...
+  init.credentials = 'include';
+  switch (endpoint) {
+    case 'menu':
+      url = '/admin-api/menu?_format=json';
+      break;
+    default:
+      break;
+  }
+
+  const data = await fetch(
```
dawehner: This is missing ```{	
         credentials: 'include',	
        },```

dawehner: Is there a reason for this mix of await and promises?

src/reducers/application.js
```diff
@@ -0,0 +1,48 @@
+import { SET_ERROR, MENU_LOADED } from '../actions/application';
+
+const initialState = {
+  error: null,
```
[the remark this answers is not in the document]
justafish: @dawehner yeah that's a really good idea. Is this something that you think should be done in this PR or can we work on something in a follow up?

src/utils/api/api.js
justafish: @dawehner how else would you write this?

src/reducers/application.js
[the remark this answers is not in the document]
dawehner: I'm happy to have a look at this in a follow up.

src/utils/api/api.js
dawehner: Couldn't you use ```const res = await fetch(
  process.env.REACT_APP_DRUPAL_BASE_URL + url,
  init,
);
const data = await res.json();
return data;
```

mattgrill:
@justafish @dawehner needs a rebase/merge with master.

PR #55, opened 2018-03-26
Try to get the router working on clicking the menu

dawehner:


src/App.js
```diff
@@ -16,32 +22,41 @@ import base from './styles/base'; // eslint-disable-line no-unused-vars
...
 const store = createStore(
-  reducers,
-  {},
-  composeWithDevTools(applyMiddleware(sagaMiddleware)),
+  combineReducers({ ...reducers, router: routerReducer }),
+  composeWithDevTools(applyMiddleware(sagaMiddleware, middleware)),
 );
 sagaMiddleware.run(actions);
 
+const withDefault = component => () => (
```
dawehner: We introduced this, as otherwise the Default component is rendered before the routes are available (they are defined inside). I don't understand why exactly it doesn't work, but this fixes it.

lauriii:
Looks good! 🥇

tedbow:
with this PR I am able to click "Permissions" and it actually load the permissions page which doesn't work with master.

src/App.js
mattgrill: @lauriii Do you have any thoughts on this?

mattgrill:
```

mattgrill: This is another way to do this, but like @dawehner pointed out, it's maybe a bit too complicated.

lauriii: I think the current approach works fine. Passing objects makes sense when you have more parameters (or you are likely to have).

mattgrill: 👏 👍

Issue #56, opened 2018-03-27
Prop type validation error on loading the menu tree


lauriii:
Menu tree resource property `subtree` is an object when there are values, but an array when the list is empty. Currently, we are only expecting for an object type of values on this property.

dawehner:
Wow, that's pretty "nice" :) Given that these keys are mostly meaningless anyway I think adapting the normalizer on the PHP side isn't the worst idea.

lauriii:
Fixed on #57

PR #57, opened 2018-03-27
Make subtree data type consistently an array

lauriii:
#56

### DIFF
--- a/src/components/06_wrappers/Default/Default.js
+++ b/src/components/06_wrappers/Default/Default.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { func, node, objectOf, shape, string } from 'prop-types';
+import { func, node, objectOf, arrayOf, shape, string } from 'prop-types';
 import { connect } from 'react-redux';
 import LoadingBar from 'react-redux-loading-bar';
 import { css } from 'emotion';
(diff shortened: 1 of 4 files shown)


src/components/06_wrappers/Default/Default.js
```diff
@@ -117,12 +117,14 @@ Default.propTypes = {
   error: string,
   menuLinks: objectOf(
```
dawehner: Could we do this on the top level too?

dawehner: Wait, isn't https://github.com/jsdrupal/drupal-admin-ui/pull/57/files#diff-8b5084ad240e61c05a90cbae6a377e86R101 doing that?

src/reducers/application.js
```diff
@@ -14,28 +14,24 @@ export default (state = initialState, action) => {
...
-                url: '/admin/people/permissions',
-              },
-            });
-          }
-          acc[key] = menuLink;
-          return acc;
-        },
-        {},
-      );
+      const menuLinks = action.payload.menuLinks.map(menuLink => {
```
dawehner: I'm glad this makes it easier here too.

PR #59, opened 2018-03-27
Create jsonapi_tables submodule to allow accesses watchdog(and other tables) via JSON API

tedbow:
Recreate the watchdog page

Right now this branch attempts to expose watchdog table entries as entities without any UI. This will allow them to be accessed via JSON API. 

This is avoid using REST View export.


dawehner:
> Right now this branch attempts to expose watchdog table entries as entities without any UI. This is avoid using REST View export.

tedbow:
@dawehner you may be right but sucks that we would have rely on views for this.

Anyways right now it shows the collections and individual GET.

To try now you have to 

1.  delete you watchdog entries  - `admin/reports/dblog/confirm`
2. drush pmu admin_ui_support -y
3. drush en admin_ui_support

tedbow:
The more I think about this more I think:

1. Yes it is weird to make watchdog entries entities
2. It very bad DX from a decoupled standpoint to have a different way to retrieve them than our jsonapi resources

We are adding jsonapi to core because it is know spec and it will make interacting Drupal (hopefully) easier for those that don't know Drupal. 

Right now core has about 20 implementations of hook_schema so that would mean at least 20 types of things that would not available via jsonapi. Some of those are strictly internal but probably if we want to build a decoupled admin ui some of those things will be very useful, like watchdog entries.

Right now the only way to expose API points easily for watchdog entries(not sure about others) is to create Views. If you are not a Drupal developer this will be very difficult. Also if we do this something will be very difficult, sorting base of query parameters and filter on any field.  Also filtering and pagination will act differently than our jsonapi resources.

We could create a custom resource for watchdog entries but then if we want to make the DX good for that resource it would make sense to make that resource behave like a jsonapi resource. It is going to need all of the things that jsonapi provides, sorting, filtering, and pagination.

Supporting  sorting, filtering, and pagination probably would not be that hard but if it doesn't behave like all other API's that Drupal provides through jsonapi we are going to have document how this resource behaves differently.

Then how would explain to javascript developer why this 1 resource acts differently then all the others? Because it is not exposing a "drupal entity"? Why would they care?

So in the end I think our *goal* should be for as many as our API resources to behave the same way. I think this current method achieves that DX goal.

How it is actually implemented in core I think can be decided later.

dawehner:
While I totally agree with you that there should be a consistent way to interact with Drupal I think nevertheless dblog entries are special:

* Any sane site out there will NOT store those in Drupal. Having to create database entries is a major performance hit
* Given that for these sites dblog will simply not be available as an endpoint
* I kind of assume that given that, it will not be common to retrieve the dblog entries.

On the other hand I had a look at ```modules/jsonapi/src/Controller/RequestHandler.php``` and I'm wondering whether ```    $resource = $this->resourceFactory($route, $current_context);``` could be sort of pluggable so that dblog could provide its own implementation.

admin_ui_support/src/ReadOnlyTableEntityStorage.php
```diff
@@ -0,0 +1,30 @@
...
+
+namespace Drupal\admin_ui_support;
+
+use Drupal\Core\Entity\EntityTypeInterface;
+use Drupal\Core\Entity\Sql\SqlContentEntityStorage;
+
+/**
+ * SQL Entity storage for entity types that wrap readonly tables.
+ */
+class ReadOnlyTableEntityStorage extends SqlContentEntityStorage {
```
dawehner: I'd expected to override ```\Drupal\Core\Entity\Sql\SqlContentEntityStorage::save``` to be overridden to ensure nothing can  ever save it

tedbow: I added this and throwing a logic exception if save() is called.

Also removed the need for WatchdogStorage class because it was only needed to handle watchdog table not having a UUID field. Move this logic to ReadOnlyTableEntityStorage

admin_ui_support/modules/jsonapi_tables/jsonapi_tables.info.yml
```diff
@@ -0,0 +1,7 @@
+name: 'JSONAPI Table'
```
dawehner: Should we mark this as experimental 🤔

admin_ui_support/modules/jsonapi_tables/src/EntityTypeTableSchemaTrait.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
+namespace Drupal\jsonapi_tables;
+
+use Drupal\Core\Entity\EntityTypeInterface;
+
+trait EntityTypeTableSchemaTrait {
```
dawehner: Nice idea!

admin_ui_support/modules/jsonapi_tables/src/ReadOnlyTableEntityStorage.php
```diff
@@ -0,0 +1,57 @@
...
+   * This prevents the module from not being able to be deleted.
+   */
+  public function hasData() {
+    return FALSE;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function save(EntityInterface $entity) {
```
dawehner: Note: we could override ```delete``` too?

admin_ui_support/modules/jsonapi_tables/src/ReadOnlyTableEntityStorage.php
```diff
@@ -0,0 +1,57 @@
...
+ * SQL Entity storage for entity types that wrap readonly tables.
+ */
+class ReadOnlyTableEntityStorage extends SqlContentEntityStorage {
+
+  use EntityTypeTableSchemaTrait;
+  /**
+   * {@inheritdoc}
+   */
+  public function loadByProperties(array $values = []) {
+    // @todo Can we add uuid to tabls to remove this hack?
```
dawehner: It feels like it could be useful to have that in dblog entries themselves. Maybe we could open up a core issue about that?

tedbow:
I have extracted this logic into jsonapi_tables which could be used to expose any table defined in `hook_schema` as a read-only entity. 
Right now it only exposes watchdog table but to expose another table you just need a class that extends `\Drupal\jsonapi_tables\Entity\ReadOnlyTableEntityBase` with a annotation. This could be changed to deriver but it might not be necessary because I don't think we want expose all tables by default.

I am interested in exporting this to contrib module at some point but probably not now until we determine this is good idea.

admin_ui_support/modules/jsonapi_tables/src/ReadOnlyTableEntityStorage.php
tedbow: Added!

admin_ui_support/modules/jsonapi_tables/jsonapi_tables.info.yml
tedbow: I would assume this whole repo is experimental right now so don't think there is need to mark this.

admin_ui_support/modules/jsonapi_tables/src/ReadOnlyTableEntityStorage.php
tedbow: I don't really see that watchdog entries need to be "Universally Unique across multiple systems"

I do think where this exposing of the tables falls down in what Drupal entities are meant to be since UUID are required for Drupal entities.

So maybe this would make incompatible with some parts of core. Workspaces? 

So I think by time we want to add this kind of functionality to core might make sense make changes to jsonapi to allow it to expose non-entity things instead of this method.

PR #60, opened 2018-03-28
Roles use Redux

mattgrill:


src/actions/application.js
```diff
@@ -39,6 +39,37 @@ function* loadMenu() {
   }
 }
 
-export default function* rootSaga() {
-  yield takeLatest(MENU_REQUESTED, loadMenu);
+export const ROLES_REQUESTED = 'ROLES_REQUESTED';
```
dawehner: Given that this is more pages specific, what about moving these to a new file, like ```roles.js``` and ```permissions.js```?

src/components/05_pages/Roles/Roles.js
```diff
@@ -1,39 +1,18 @@
...
-import Error from '../../02_atoms/Error/Error';
 
 const Roles = class Roles extends Component {
-  state = {
-    roles: {},
-    loaded: false,
-  };
-  componentDidMount() {
-    this.cancelFetch = makeCancelable(this.fetchData());
+  componentWillMount() {
```
dawehner: I think we should still cancel using https://redux-saga.js.org/docs/api/#canceltask

mattgrill: @dawehner I had a look at this stuff. It sort of looks like whatever xhr library you are using needs to support some sort of cancel/abort method, which I don't think `fetch` does.

dawehner: https://stackoverflow.com/questions/46429081/react-redux-saga-event-channel-cancel seems to do that (see first answer)

dawehner: After reading https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#fetching-external-data I think we should continue using componentDidMount

mattgrill: Yes, sorry. This was an oversight on my part.

src/App.js
```diff
@@ -39,7 +39,7 @@ const withDefault = component => () => (
 
 class App extends Component {
   componentDidMount() {
-    history.replace('/');
+    // history.replace('/');
```
lauriii: Should we just delete this? I never understood why we would want to always redirect to the frontpage on the initial page load.

lauriii: I just read the other PR from @dawehner and now I understand why we need this.

mattgrill: However, it's commented out here because it breaks routing right now.

mattgrill: If you just run this from the dev server, and try and visit a page, you won't actually get anything to load.

mattgrill: broken-navigation

PR #66, opened 2018-03-31
Update README.md

justafish:
This PR adds instructions for contributing to the repository using the fork model because we probably shouldn't be giving write access to the repository to everyone in the entire world 😅


README.md
```diff
@@ -1,6 +1,12 @@
+# vfancy
```
dawehner: Let's call it drupal admin ui :)

PR #68, opened 2018-04-02
Provide some initial test coverage

lauriii:


src/App.test.js
```diff
@@ -1,9 +1,9 @@
...
-import ReactDOM from 'react-dom';
+import { shallow } from 'enzyme';
+
 import App from './App';
 
 it('renders without crashing', () => {
-  const div = document.createElement('div');
-  ReactDOM.render(<App />, div);
-  ReactDOM.unmountComponentAtNode(div);
+  const root = shallow(<App />);
```
dawehner: I'm curious, what happens with the IO that should be fired here?

src/components/05_pages/Roles/Roles.test.js
```diff
@@ -0,0 +1,28 @@
...
+    // Ensure that task is cancelled on unmount.
+    root.unmount();
+    expect(cancelTask).toHaveBeenCalled();
+  });
+
+  it('loading animation is shown before roles have been loaded', () => {
+    const root = shallow(
+      <Roles cancelTask={() => {}} requestRoles={() => {}} roles={null} />,
+    );
+    expect(root.find(LoadingBar).length).toEqual(1);
```
dawehner: That is super nice!

src/App.test.js
lauriii: This actually only renders the main level so we don't have to worry about that as part of this. We should solve problems related to that in a follow-up issue

dawehner: Ah gotcha, nevermind :)

src/routes.js
```diff
@@ -1,5 +1,5 @@
 import Permissions from './components/05_pages/Permissions/Permissions';
-import Roles from './components/05_pages/Roles/Roles';
+import Roles from './components/05_pages/Roles';
```
dawehner: Works for me!

mattgrill:
Can we add a task to Circle CI to run these?

src/components/06_wrappers/ErrorBoundary/ErrorBoyndary.test.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
```
larowlan: filename has a typo here

dawehner: I fixed that in https://github.com/jsdrupal/drupal-admin-ui/pull/73

PR #69, opened 2018-04-02
Document available cli commands

lauriii:


lauriii:
We have one command for prettier which I'm not sure what is the difference for `yarn run test:lint:fix`. Does anyone else know what is the difference between those two?

dawehner:
> We have one command for prettier which I'm not sure what is the difference for yarn run test:lint:fix. Does anyone else know what is the difference between those two?

That one uses eslint fixing, which is partially a sub and superset of what prettier does :) I think we should combine the two though, even it is slower to run the eslint one.

PR #70, opened 2018-04-02
Initial experiment: Generate configuration based upon configuration schema

dawehner:
As outlined on the initial issue on drupal.org we want to enable people to build excellent admin UIs, for special things like the field UI. On the other hand there are many pages all over the place, which are way more simple, for example the site information form: ```admin/config/system/site-information``` .

In order to not require people to implement a lot of those forms manually too, we came up with the idea to autogenerate forms in JS.

For that there are a lot of metadata information needed. For now we are building based upon config schema, as this is what defines configuration. Let's get this experiment started.

In the future we might build something upon json schema, but there are a lot of additional information in configuration schema, see https://www.drupal.org/project/drupal/issues/2949888

## How to test it

* Generate the configuration schema file into the drupal-admin-ui folder:
```
cd drupal-installation
drush scr /drupal-admin-ui/extractConfigSchema.php  > /drupal-admin-ui/src/configSchema.json
```
* Goto ```/admin/config/system/site-information``` in the app

## Missing points

* We don't invoke configuration validation yet on the server site (TODO add issue)
* We don't how configuration validation yet client side
* Note: We need actual validation for this form in Drupal: https://www.drupal.org/project/drupal/issues/2952038
* We should extract the form components into our own component library and just use them yet
* Do we want to actually use ```react-jsonschema-form``` one day?
* There is no clean history in this branch :)

screen shot 2018-04-02 at 10 29 24


.gitignore
```diff
@@ -21,3 +21,5 @@
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+
+.idea
```
dawehner: Does anyone care about this?

src/reducers/application.js
```diff
@@ -79,6 +81,18 @@ export default (state = initialState, action) => {
         roles,
       };
     }
+    case SIMPLE_CONFIG_LOADED: {
```
dawehner: It is a bit weird to use redux/saga for initial loading but then local state for the actual form.

dawehner:
Let's close this for now

PR #73, opened 2018-04-02
Allow to filter the menu

dawehner:


src/components/06_wrappers/Default/Default.js
```diff
@@ -138,15 +153,26 @@ Default.propTypes = {
...
 
-const mapStateToProps = state => ({
-  message: state.application.message || null,
-  menuLinks: state.application.menuLinks || {},
-});
+const mapStateToProps = state => {
+  return {
+    message: state.application.message || null,
+    menuLinks:
+      state.application.filteredMenuLinks.length > 0
```
lauriii: For me it seems like a bit strange UX that we show all results if none is found with the given string. How about we change this to just check if `filterString` exists?

src/reducers/application.js
```diff
@@ -72,6 +97,21 @@ export default (state = initialState, action) => {
         menuLinks,
       };
     }
+    case MENU_FILTERED:
```
lauriii: Is there a particular reason why we have to store the computed data in the store? If not, we could probably do this on run time while getting the data. Here's link for the redux docs which should provide some idea what could be done https://redux.js.org/recipes/computing-derived-data

src/components/06_wrappers/Default/Default.js
dawehner: Oh good point! :+1:

src/reducers/application.js
dawehner: @lauriii I followed the approach @mattgrill was doing on the permissions page. I think having less state aka. computing the values as you go, is the better approach. If there are performance issues with it, we could always come back to it. 👍

mattgrill: Yeah. The permissions page needs some work right now since it's still not using the saga/redux pattern. At that migration point it would be nice to reduce the JSON API payload to just the items we need.

dawehner: Did we agreed on using redux for storing the filter string for now?

lauriii: If it's not too much work, I think it would be better to move the string to the local state :)

dawehner:
Oh, I guess I changed it too quickly. I'll probably address this before tonight.

dawehner:
The filtering is back.

PR #75, opened 2018-04-05
Improve styles

lauriii:


dawehner:
First impression: Way nicer! Using open sans makes this look really nice.

src/components/05_pages/Permissions/Permissions.js
```diff
@@ -221,30 +224,36 @@ const Permissions = class Permissions extends Component {
...
+                  (this.state.working && this.state.changedRoles.length)
+                }
+              >
+                Save
+              </button>
+            </div>
+          )}
+        </Sticky>
         <Table>
           <THead
```
dawehner: A future idea: It would be nice to get sticky headers for the table header too. I tried to get it working, but another ```<Sticky>``` header, even with an additional 80 pixels of offset didn't work.

src/components/06_wrappers/Default/Default.js
```diff
@@ -69,13 +77,12 @@ styles = {
...
   main: css`
-    padding: 1rem;
-    margin-top: 100px;
+    padding: 7rem 2rem 3rem;
     height: 100%;
     background: #ffffff;
   `,
   burgerButton: css`
-    position: fixed;
+    position fixed;
```
dawehner: Is there a reason why the common was dropped?

PR #77, opened 2018-04-05
Bug fixes on the permissions page (list below)

dawehner:
* Ignores casing on search completely
* Ensure saving the permissions works


src/components/05_pages/Permissions/Permissions.js
```diff
@@ -13,6 +13,14 @@ import {
 import Loading from '../../02_atoms/Loading/Loading';
 import { Table, TBody, THead } from '../../01_subatomics/Table/Table';
 
+export const filterPermissions = (input, permissions) =>
+  permissions.filter(
+    ({ title, description, provider, provider_label: providerLabel }) =>
+      `${title}${description}${provider}${providerLabel}`
+        .toLowerCase()
+        .includes(input.toLowerCase()),
```
mattgrill: Instead of lowercasing the input on each loop wouldn't it be easier to lowercase it when it's passed in?

dawehner: For me at least encoding as much as possible within this one magic helps to understand its complete behavior (being case insensitive).

PR #80, opened 2018-04-12
Ensure the redirect also works on the build environment

dawehner:


dawehner:
I'm merging it right now as the app is broken completely otherwise and that's bad for people here at the sprint table.

PR #86, opened 2018-04-13
Add controller to return dblog types

tedbow:
This should likely be JSON RPC plugin if we build that system.

path: /admin-ui-support/dblog-types?_format=json

response:
`[
"cron",
"page not found",
"php",
"system",
"user"
]`


admin_ui_support/src/Controller/DbLogTypes.php
```diff
@@ -0,0 +1,55 @@
...
+   * Gets current dblog types.
+   */
+  public function get() {
+    $levels = [];
+
+    /* @todo Should the route to this controller be created dynamically or
+     * should this module be dependent on dblog? How does the React app handle
+     * routes for modules that aren't enabled.
+     */
+    if ($this->moduleHandler->moduleExists('dblog')) {
```
dawehner: Well, I think this is totally fine till we maybe have a RPC

dawehner:
@tedbow Do you want trying to setup a new branch with the current content? Maybe this fixes circleci

Issue #85, opened 2018-04-13
Implement simple content listing page


dawehner:

tedbow:
To support this I am looking what it would take for create `/jsonapi/node` resource that would allow collections of nodes(another entity types) across bundles.

It doesn't look like the *jsonapi* module itself would introduce this https://www.drupal.org/project/jsonapi/issues/2956414

Seems like it would be possible with new resource type. 

My idea would be figure who to expose just the basefields for filtering, sorting. 
So you could still search by title, type, date, etc.

dawehner:
What confuses me personally is the following: Entity queue is totally able to handle cross bundle queries.

tedbow:
This https://www.drupal.org/project/entityqueue or \Drupal\Core\Config\Entity\Query\Query ?

Assuming the 2nd. Yeah I think it is more design decision for entity api.

PR #87, opened 2018-04-13
Content page

samtimalsina:
Issue: #85 

This adds a basic content page at /admin/content, but only lists articles at the moment.

screen shot 2018-04-13 at 1 04 01 pm

### DIFF
--- a/src/utils/api/api.js
+++ b/src/utils/api/api.js
@@ -8,6 +8,8 @@ async function api(endpoint, init = {}) {
     case 'roles':
       url = '/jsonapi/user_role/user_role';
       break;
+    case 'content':
+      url = '/jsonapi/node/article';
     default:
       break;
   }
--- a/src/utils/api/api.js
+++ b/src/utils/api/api.js
@@ -10,6 +10,7 @@ async function api(endpoint, init = {}) {
       break;
     case 'content':
       url = '/jsonapi/node/article';
+      break;
     default:
       break;
   }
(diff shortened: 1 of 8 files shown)


src/utils/api/api.js
```diff
@@ -8,6 +8,9 @@ async function api(endpoint, init = {}) {
     case 'roles':
       url = '/jsonapi/user_role/user_role';
       break;
+    case 'content':
```
dawehner: So yeah we miss the feature to fetch nodes from multiple bundles here :(

justafish:
Thanks @samtimalsina!

This route will need adding to [admin_ui_support.routing.yml](https://github.com/jsdrupal/drupal-admin-ui/blob/master/admin_ui_support/admin_ui_support.routing.yml) so that the redirect from Drupal works

The linting tests also need fixing

This is great! 🎉

src/utils/api/api.js
tedbow: #96 would allow this!

src/utils/api/api.js
```diff
@@ -8,6 +8,9 @@ async function api(endpoint, init = {}) {
...
+      url = '/jsonapi/node/article';
```
tedbow: Now that https://github.com/jsdrupal/drupal-admin-ui/pull/96 was merged you should be able to change this to `/json/node` to query all node entities.

mattgrill:
@samtimalsina I've closed this, and taken your work (co-authored a commit) and added upon that work to request all the content from all the bundles. It's all in #101

PR #89, opened 2018-04-14
Fix the redirect for real

dawehner:


dawehner:
@rocketeerbkw Thank you for giving this a proper testing. I've made some changes and all the cases you described above seems to work properly now.

src/App.js
```diff
@@ -63,6 +54,8 @@ class App extends Component {
                   key={route}
                 />
               ))}
+              <Route path="/vfancy" component={withRouter(InitialRedirect)} />
+              <Route path="/vfancy/" component={withRouter(InitialRedirect)} />
```
mattgrill: I think you can make trailing slashes optional with something like `(\?)`?

Issue #21, opened 2018-03-14
Add a favicon


dawehner:
What about using the logo ◀️👋▶️ as favicon?

Tagged as novice task / help wanted

justafish:
YES

lauriii:
https://github.com/jsdrupal/drupal-admin-ui/pull/90 fixed this 👍

PR #90, opened 2018-04-15
Upate favicon

zweishar:
Updates favicon to image suggested in #21 

Uses "Pale Gray" from [Drupal style guide](https://www.drupal.org/drupalorg/style-guide/colors) to fill empty space.


dawehner:
Thank you @zweishar ! I like the level of detail by choosing one of the colours.
I'm curious whether we should generate more favicons, see https://stackoverflow.com/questions/4014823/does-a-favicon-have-to-be-32x32-or-16x16 
Maybe running your high res version through https://realfavicongenerator.net would be nice. Any thoughts?

justafish:
They can also go into a web manifest file: https://developer.mozilla.org/en-US/docs/Web/Manifest

zweishar:
Thanks all, those sounds like good suggestions. I will update this PR with some additional favicon sizes and update the web manifest accordingly.

zweishar:
After reviewing https://realfavicongenerator.net and https://serverless-stack.com/chapters/add-app-favicons.html I believe this setup should give us an icon on most devices / browsers.

I also brought the edges in and rounded the corners to make it look better when used as an icon on other devices.

justafish:
Thank you @zweishar ! 🙏

PR #91, opened 2018-04-15
Move role/permission fetching and posting into api && make the module compatible with jsonapi 1.15

dawehner:
Note: I tried to play around with using redux/sagas for this page, but I thought this PR would be a smaller change.

### DIFF
--- a/admin_ui_support/src/Controller/JsonApiRequestHandler.php
+++ b/admin_ui_support/src/Controller/JsonApiRequestHandler.php
@@ -22,15 +22,18 @@ class JsonApiRequestHandler extends RequestHandler {
    * {@inheritdoc}
    */
   protected function resourceFactory(ResourceType $resource_type) {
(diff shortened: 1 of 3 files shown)


admin_ui_support/src/Controller/JsonApiRequestHandler.php
```diff
@@ -19,22 +21,16 @@ class JsonApiRequestHandler extends RequestHandler {
...
-        $this->currentContext,
-        $this->fieldTypeManager,
-        $this->linkManager,
-        $this->resourceTypeRepository
-      );
-    }
-    // If the entity is not a config entity default to the parent.
-    return parent::resourceFactory($route);
+  protected function resourceFactory(ResourceType $resource_type) {
+    $resource = new ConfigEntityResource(
```
dawehner: Oh this IF is still needed

PR #93, opened 2018-04-16
Add controller to return dblog types (try 2)

tedbow:
Retrying: https://github.com/jsdrupal/drupal-admin-ui/pull/86

There was CI problem that @dawehner suggested could be fixed by creating new PR

~

This should likely be JSON RPC plugin if we build that system.

Request: `/admin-ui-support/dblog-types?_format=json`
Response:
```json
[
	"cron",
	"page not found",
	"php",
	"system",
	"user"
]
```


admin_ui_support/src/Controller/DbLogTypes.php
```diff
@@ -0,0 +1,55 @@
...
+    $levels = [];
+
+    /* @todo Should the route to this controller be created dynamically or
+     * should this module be dependent on dblog?
+     * How does the React app handle routes for modules that aren't enabled?
+     */
+    if ($this->moduleHandler->moduleExists('dblog')) {
+      $levels = array_values(_dblog_get_message_types());
+    }
+    return new JsonResponse($levels);
```
dawehner: Should we cache this response?

tedbow: I am not sure what we would cache it on.  `type` is just an arbitrary string so we would have to check after every message is written to see if there were new types.

tedbow: Also I don't think Drupal itself caches query so I don't think we should.

dawehner: Good point, yeah, let's not cache it :)

PR #95, opened 2018-04-17
Add configuration option to disable redirecting routes.

tedbow:
If you are using the React totally separately from the Drupal site you may want to disable redirecting paths from Drupal to the app.

Provides a new form at `/admin/config/admin_ui_support/settings` to enable/disable redirecting.  When the module is first enabled it will redirect.

### DIFF
--- a/admin_ui_support/src/Form/SettingsForm.php
+++ b/admin_ui_support/src/Form/SettingsForm.php
@@ -1,0 +1,93 @@
+<?php
+
+namespace Drupal\admin_ui_support\Form;
+
+use Drupal\Core\Config\ConfigFactoryInterface;
+use Drupal\Core\Form\ConfigFormBase;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Routing\RouteBuilderInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Settings form for JS Drupal App integration.
+ */
+class SettingsForm extends ConfigFormBase {
+
+  /**
+   * The router rebuilder.
+   *
+   * @var \Drupal\Core\Routing\RouteBuilderInterface
+   */
+  protected $routerRebuilder;
+
+  /**
+   * Constructs a \Drupal\system\ConfigFormBase object.
+   *
+   * @param \Drupal\Core\Config\ConfigFactoryInterface $config_factory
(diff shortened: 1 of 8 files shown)


admin_ui_support/src/Routing/RouteSubscriber.php
```diff
@@ -11,10 +12,27 @@
...
+   */
+  public function __construct(ConfigFactoryInterface $config_factory) {
+    $this->configFactory = $config_factory;
+  }
+
   /**
    * {@inheritdoc}
    */
   protected function alterRoutes(RouteCollection $collection) {
+    if (!$this->configFactory->get('admin_ui_support.settings')->get('redirect_related_routes')) {
```
dawehner: Did you thought about making this distinction on runtime, so you don't need to rebuild the router for that?

tedbow: Well we are already replacing the controller so we would have lost the original controller.

I don't think this is something that would happen very often at all. So once the routes are rebuilt there is not effect at runtime which seems better for something that would switch very infrequently.

dawehner: Yeah fair, let's not worry for now :)

admin_ui_support/src/Form/SettingsForm.php
```diff
@@ -0,0 +1,93 @@
...
+   */
+  public function submitForm(array &$form, FormStateInterface $form_state) {
+    parent::submitForm($form, $form_state);
+
+    $redirect_related_routes = $form_state->getValue('redirect_related_routes');
+    if ($this->config('admin_ui_support.settings')->get('redirect_related_routes') !== $redirect_related_routes) {
+      $this->config('admin_ui_support.settings')
+        ->set('redirect_related_routes', $redirect_related_routes)
+        ->save();
+      $this->routerRebuilder->rebuild();
```
dawehner: Instead of directly rebuilding the router we should make it lazy, by calling to : ```setRebuildNeeded``` instead.

admin_ui_support/tests/src/Functional/RedirectRoutesTest.php
```diff
@@ -0,0 +1,56 @@
+<?php
```
dawehner: Nice test!

PR #96, opened 2018-04-18
Jsonapi collections across bundles to allow creating `admin/content`

tedbow:
*work in progress*

#85 would require fetching all nodes not limited by bundle. All of the jsonapi's resource are limited by bundle with the path `/jsonapi/[entity_type]/[bundle]`

This PR entity level collection-only resources at `/jsonapi/[entity_type]`

I created a new module `jsonapi_support` which now combines the previous functionality added in #29 to allow updating config entities with the new functionality for the new resources.

Because the way that `jsonapi_support` works it would not be compatible with the `jsonapi_extras` module. As far as know we don't need `jsonapi_extra` for the `drupal-admin-ui` React app.


tedbow:
Added basic tests and fix removing POST method.

PR #101, opened 2018-04-24
Content page

mattgrill:
A content page.

- [x] Filters
- [x] Sorting.


mattgrill:
Alright, 👍 from me. I'll merge this now and when we get to the point of having some custom styles for the material-ui components we can start the wrapping of the components with our namespace/styles.

src/actions/application.js
```diff
@@ -48,6 +48,37 @@ function* loadMenu() {
   }
 }
 
+/**
+ * Gets all available content types.
+ */
+export const CONTENT_TYPES_REQUESTED = 'CONTENT_TYPES_REQUESTED';
```
dawehner: It feels really weird to put these reducers into application.js. Is there a reason this is not moved to a admin/content specific page?

src/utils/api/api.js
```diff
@@ -36,6 +36,14 @@ async function api(
     case 'permissions':
       url = '/admin-api/permissions?_format=json';
       break;
+    case 'content':
```
dawehner: It is weird that there is 'content' and 'nodeType' :)

src/components/05_pages/Content/Content.js
```diff
@@ -0,0 +1,218 @@
...
+            <MenuItem value="">
+              <em>Any</em>
+            </MenuItem>
+            <MenuItem value="published">Published</MenuItem>
+            <MenuItem value="unpublished">Unpublished</MenuItem>
+          </Select>
+        </FormControl>
+      </div>
+      <Paper>
+        <Table>
```
lauriii: Should we show a loading animation at least on the initial page load?

src/components/05_pages/Content/Content.js
```diff
@@ -0,0 +1,218 @@
...
+              ].map(item => <TableCell key={item}>{item}</TableCell>)}
+            </TableRow>
+          </TableHead>
+          <TableBody>
+            {this.props.nodes.map(
+              ({ type, attributes: { changed, nid, status, title } }) => (
+                <TableRow key={nid}>
+                  <TableCell>{title}</TableCell>
+                  <TableCell>{this.props.contentTypes[type].name}</TableCell>
+                  <TableCell />
```
lauriii: Should we remove this column for now and create a follow-up issue to add it?

src/components/05_pages/Content/Content.js
```diff
@@ -0,0 +1,218 @@
...
+                    </Button>
+                    <Button
+                      variant="fab"
+                      disabled
+                      aria-label="delete"
+                      className={styles.button}
+                      component={Link}
+                      to={`/node/${nid}/delete`}
+                    >
+                      <DeleteIcon />
```
lauriii: Should we link this to the delete page for now?

src/components/05_pages/Content/Content.js
justafish: @lauriii I removed the LoadingBar from this page as I think we should move it to the Default template, but didn't want to start exploring that here

lauriii: @justafish makes sense! Let's not block this PR with that, but we should open an issue for that anyway! 🙂

src/components/05_pages/Content/Content.js
mattgrill: This is the user?

src/actions/application.js
justafish: @dawehner exactly :+1:

src/components/05_pages/Content/Content.js
```diff
@@ -157,7 +157,6 @@ class Content extends Component {
               {[
                 'Title',
                 'Content Type',
-                'Author',
```
mattgrill: I'll take care of adding the author in a followup. It shouldn't be too difficult.

src/App.js
```diff
@@ -35,29 +37,40 @@ const store = createStore(
...
+  <JssProvider jss={jss} generateClassName={generateClassName}>
+    <ErrorBoundary>
+      <Provider store={store}>
+        <ConnectedRouter history={history}>
+          <Switch>
             <Route
-              path={route}
-              component={withDefault(withRouter(routes[route]))}
-              key={route}
+              exact
```
dawehner: In order to replicate the behaviour of Drupal right now we would need to check "access content overview" on this route. Of course we can do this later, but I think access checking on routes, not just endpoints, is something we should keep in mind somehow.

PR #102, opened 2018-04-25
Test Drupal + Fix tests

mattgrill:
- Add PHP/Drupal test steps to CircleCI
- Fix tests so they are correct.


.circleci/config.yml
```diff
@@ -13,6 +21,54 @@ aliases:
...
+            - v4-dependencies-{{ checksum "custom/admin_ui_support/composer.json" }}
+            - v4-dependencies-
+      - run:
+          working_directory: /var/www/html
+          command: |
+            ./update-dependencies.sh $CIRCLE_PROJECT_REPONAME
+      # Add JSON API
+      - run:
+            working_directory: /var/www/html
+            command: /usr/local/bin/composer require drupal/jsonapi:^1.16
```
deviantintegral: I think this can be removed, as you have (properly) added jsonapi as a requirement in composer.json below.

.circleci/config.yml
```diff
@@ -1,4 +1,12 @@
 aliases:
+  - &drupal_defaults
+    docker:
+      - image: andrewberry/drupal_tests:0.3.4
+      - image: selenium/standalone-chrome-debug:3.7.1-beryllium
+      - image: mariadb:10.3
```
dawehner: why are we using mysql and not sqlite?

mattgrill: @davidhwang This is just copied directly from @deviantintegral :)

mattgrill: Trying for the least friction to get testing working :)

dawehner: sure, no worries

deviantintegral: So we actually started with sqlite, and ran into a few issues and [switched to mariadb](https://github.com/deviantintegral/drupal_tests/pull/5).

1. Behat tests would sporadically fail if there were enough concurrent requests with sqlite lock exceptions.
1. There's a few contributed modules out there (we ran into a module depending on dynamic_entity_reference) that depended on mariadb-only features like triggers. With a mariadb container basically being free under circleci we've yet to see any issues since switching.

dawehner: One nice reason to use sqlite would be that we tell people to test out drupal-admin-ui using sqlite too. This way we would be more in line with the actual environment people use.

.circleci/config.yml
dawehner: That seems to make sense

admin_ui_support/src/Plugin/rest/resource/PermissionsCollectionResource.php
```diff
@@ -105,7 +105,7 @@ public function get() {
       $permissions = $this->permissionHandler->getPermissions();
       foreach ($permissions as $id => $permission) {
         $permissions[$id]['id'] = $id;
-        $permissions[$id]['provider_label'] = $this->moduleExtensionList->getName($permissions[$id]['provider']);
+        $permissions[$id]['provider_label'] = $this->moduleHandler->getName($permissions[$id]['provider']);
```
dawehner: Why do have these changes? I think this is wrong given getName is not meant to be the actual API.

.circleci/config.yml
mattgrill: Just a followup on this. I tried to remove this and JSON API was not actually installed.

admin_ui_support/src/Plugin/rest/resource/PermissionsCollectionResource.php
mattgrill: @tedbow

dawehner:
In order for this to land I think we need to be able to test with 8.6.x-dev. Otherwise we end up with a non realistic testing environment.

mattgrill:
I guess this might produce some build/test inconsistencies.

admin_ui_support/src/Plugin/rest/resource/PermissionsCollectionResource.php
tedbow: Yeah I think I was trying to get this to work with 8.5.x  since the 'extension.list.module' service doesn't exist in this branch.

Not sure if we care about 8.5.x core for this work.

mattgrill:
I hate github.

PR #103, opened 2018-04-25
Remove cross bundle collections for entity types that don't support bundles

tedbow:
\Drupal\jsonapi_support\ResourceType\ResourceTypeRepository was not checking if the entity type support bundles. 

Because this resource type did not need a bundle it was using entity_type_id for bundle name.

But because of the ids for resources are created it was replacing resources provided by jsonapi. 

For instance there was existing resource that `node_type--node_type` that had the path `jsonapi/node_type/node_type`.
This was being removed. and replace with a resource with a path `jsonapi/node_type`.

When creating a new `\Drupal\jsonapi_support\ResourceType\CrossBundlesResourceType` now it is using '__all' for bundle. Therefore if bundle name is used that is the same as the entity type it will also not replace resource.

** Testing
This broke the permissions page in the React app because there was resource at `jsonapi/user_role/user_role` but no tests failed. Which means we don't have integration tests for this.


dawehner:
Let's close this now that we decided to focus on the content author.

PR #106, opened 2018-04-29
Fix API endpoints for user roles

zweishar:
The API endpoints for Role related pages seem to have gotten mangled at some point. Currently on the master branch my Role listing page does not work, and I'm unable to save the role specific permission page. This update resolves that for me.

screen shot 2018-04-29 at 11 12 36 am


dawehner:
Oh interesting @tedbow  might be interested in this.

PR #115, opened 2018-06-05
Human readable time display.

mattgrill:
🎉


src/actions/content.js
```diff
@@ -29,7 +29,12 @@ function* loadContent(action) {
     yield put(showLoading());
 
     const queryString = {
-      sort: 'changed',
+      sort: {
```
mattgrill: This didn't actually work anymore because JSON API changed behind the scenes.

src/components/05_pages/Content/Content.js
```diff
@@ -169,7 +170,9 @@ class Content extends Component {
                   <TableCell>
                     {(status && 'Published') || 'Unpublished'}
                   </TableCell>
-                  <TableCell>{changed}</TableCell>
+                  <TableCell>
+                    {format(new Date(changed * 1000), 'MM/DD/YYYY - HH:mm')}
```
dawehner: I'm wondering why we are not using something like: ```var options = { weekday: 'long', year: 'numeric', month: 'long', day: 'numeric' };
console.log(new Intl.DateTimeFormat('de-DE', options).format(date));``` That way we don't need a library, while we also get localization based upon the browser for free.

lauriii:
Let's open issue for the timezone issue. This already improves the situation whole bunch 🌮

PR #123, opened 2018-06-06
Users for content listing.

mattgrill:


src/components/05_pages/Content/Content.js
```diff
@@ -155,17 +159,47 @@ class Content extends Component {
...
+                type,
+                attributes: { changed, nid, status, title },
+                relationships,
+              }) => (
                 <TableRow key={nid}>
                   <TableCell>{title}</TableCell>
                   <TableCell>{this.props.contentTypes[type].name}</TableCell>
+                  <TableCell>
+                    {this.props.includes['user--user'][
+                      relationships.uid.data.id
```
lauriii: This is kind of hard to read when we split this on multiple lines. Is this how our coding standards suggest us to do this?  🤔

src/reducers/content.js
```diff
@@ -9,6 +9,11 @@ export default (state = initialState, action) => {
     case CONTENT_LOADED: {
       return {
         ...state,
+        includes: action.payload.contentList.included.reduce((acc, cur) => {
```
lauriii: Maybe we could add a comment to describe what we are doing and/or rename the variables. I had to actually go with debugger and look what we are doing here to understand this.

src/components/05_pages/Content/Content.js
dawehner: I guess we could increase the maximum line length in prettier. For me though it is not unreadable, but maybe that's simply because I'm quite used to prettier output.

src/reducers/content.js
dawehner: +1 for renaming "cur" variables. How about using "node" here?

src/components/05_pages/Content/Content.js
mattgrill: I don't really have an issue with this. Prettier is sometimes a bit of pain, but at least it makes things consistent.

PR #124, opened 2018-06-06
Sort table contents. 

mattgrill:


src/components/05_pages/Content/Content.js
```diff
@@ -155,8 +162,37 @@ class Content extends Component {
...
+                ({ key, label, sortable }) =>
+                  sortable ? (
+                    <TableCell key={key}>
+                      <TableSortLabel
+                        direction={
+                          this.state.sort.name === key
+                            ? this.state.sort.direction.toLowerCase()
+                            : undefined
+                        }
+                        active={(this.state.sort.name === key && true) || false}
```
dawehner: This one is quite weird. Isn't that the same as ```(this.state.sort.name === key```?

src/actions/content.js
```diff
@@ -24,19 +24,21 @@ function* loadContent(action) {
...
-          path: 'changed',
-          direction: 'DESC',
-        },
-      },
       filter: {},
     };
+
+    if (sort) {
+      const { name, path, direction } = sort;
+      queryString.sort = { [`sort-${name}`]: { path, direction } };
```
wimleers: This is Drupal-specific sorting syntax. For simple sorts like these, I'd recommend sticking to what the spec provides: http://jsonapi.org/format/#fetching-sorting

mattgrill: This is never going to be used outside of connecting to Drupal.

dawehner: Well, changing to ```queryString.sort = [`${direction === 'DESC' ? '-' : ''}${direction}`]``` should be easy?

src/components/05_pages/Content/Content.js
```diff
@@ -79,11 +80,17 @@ class Content extends Component {
   state = {
     contentTypes: [],
     status: null,
+    sort: { path: 'changed', direction: 'DESC' },
```
lauriii: We probably should use the redux state for storing the sort settings since it's relevant information related to the data.

mattgrill: My only concern is that this information currently is only used for this page. Keeping it in Redux adds complexity that we don't currently need.

dawehner: Well with the same argument we could argue that the content should not be in the redux store either :)

mattgrill: Yeah. I agree. I think maybe we should make a _followup_ to try and decide what types of information we store in redux vs. what is kept in the component.

Issue #122, opened 2018-06-06
Add the "add content" button to content list


lauriii:
The current "add content" button is added using local actions but we could probably for now just hard code the button. I think we could use the pattern from here https://material-ui.com/demos/buttons/#floating-action-buttons for this.

PR #130, opened 2018-06-10
Add button added(#122)

iberdinsky-skilld:
Hello! 
Added 'add' button. In fixed right/top position. but not sure it should be there. Please review

### DIFF
--- a/src/components/05_pages/Content/Content.js
+++ b/src/components/05_pages/Content/Content.js
@@ -23,6 +23,7 @@ import TableHead from '@material-ui/core/TableHead';
 import TableRow from '@material-ui/core/TableRow';
 
 import Button from '@material-ui/core/Button';
+import AddIcon from '@material-ui/icons/Add';
 import EditIcon from '@material-ui/icons/Edit';
 import DeleteIcon from '@material-ui/icons/Delete';
 
@@ -33,6 +34,12 @@ const styles = {
   root: css`
     display: flex;
     flex-wrap: wrap;
+  `,
+  addButton: css`
+    margin: 0.5rem;
+    position: fixed;
+    right: 0;
+    top: 0;
   `,
   formControl: css`
     margin: 0.5rem;
@@ -154,6 +161,17 @@ class Content extends Component {
               <MenuItem value="unpublished">Unpublished</MenuItem>
             </Select>
           </FormControl>
+
+          <Button
+            variant="fab"
+            color="primary"
+            aria-label="add"
+            className={styles.addButton}
+            component={Link}
+            to="/node/add"
+          >
+            <AddIcon />
+          </Button>
         </div>
 
         <Table>


src/components/05_pages/Content/Content.js
```diff
@@ -34,6 +35,12 @@ const styles = {
     display: flex;
     flex-wrap: wrap;
   `,
+  addButton: css`
+    margin: 0.5rem;
+    position: fixed;
+    right: 0;
+    top: 0;
```
lauriii: Is there a specific reason for placing the add button to the top right corner? It seems like material UI is placing it in the bottom right corner in [their demo](https://material-ui.com/demos/buttons/#floating-action-buttons).

lauriii: Sorry, I just noticed you mentioned in the PR summary that you are not sure if it should be in the top right corner. 🙈In that case I probably would prefer to follow the material UI demo and move it to the bottom right corner. 🙂

iberdinsky-skilld: i found that bottom position of button not really obvious here. 
https://gyazo.com/fd38420bdc8c5ab346ecde6191a8d639

iberdinsky-skilld: maybe better to change 'edit' and 'delete' buttons variant? to have add button more 'special'
https://gyazo.com/66dd22b02ea1ed126d06e3e37ef8268c

lauriii: I think that is a good point. 👍 What do you think about using this pattern instead: https://material-ui.com/demos/buttons/#buttons-with-icons-and-label for the table rows?

ckrina: 100% agree with @iberdinsky-skilld suggestion. The ones in the list shouldn't use the rounded style but the plain icon because otherwise they get too much attention and compete with the main action. Unless there's an accessibility issue with using icons without a label I'd use them alone.

finnsky:
added new variant for Edit and Delete buttons:
https://gyazo.com/49904ea5e443fb9e3aae4107b28966d6
Looks better imo. but maybe one more thing should be done:
https://material-ui.com/demos/buttons/#sizes
With size small they not take lot of user attention.

iberdinsky-skilld:
added new variant for Edit and Delete buttons: 
https://gyazo.com/49904ea5e443fb9e3aae4107b28966d6
But also i think their size should be changed to small. cause now they take too much user attention.
with small size(not commited): https://gyazo.com/81fe6af110f853a8eeb663e638cbee3c

ckrina:
Sorry I didn't explained it properly: what is better for the actions in the list is the [Icon buttons](https://material-ui.com/demos/buttons/#icon-buttons) as you had here: https://gyazo.com/66dd22b02ea1ed126d06e3e37ef8268c because the Floating button gets most of the attention.

src/components/05_pages/Content/Content.js
eigentor: Definitely. Especially the red delete button dominates everything when using the rounded style.

lauriii:
Looks good for me! Thanks for making the changes 🙂

Pls:
New contributor here. Would you guys consider making buttons inline in table cell like this? I think it's better UX this way and more compacted rows.

react_app

lauriii:
@Pls welcome! 👋 I think that's what we ended up building in the end! Let me know if there's still something that you believe we could improve 🙂

Pls:
Thanks for warm welcome, @lauriii! Will do my best 😉

> I think that's what we ended up building in the end!

Can you clarify what you mean here. You mean you already done that fix or plan to do it? I think this is not implemented yet (no visible commits or PR related to that). So, maby I should just create another PR for that little improvement? 😉

lauriii:
@Pls I think I misunderstood your feedback on the first time. Is this about how the buttons are shown on a narrow window?

Pls:
@lauriii Allright, thanks! Will open up new PR soon ;)

PR #134, opened 2018-06-12
Show an error message if you messed up configuring this app

justafish:
If you set `REACT_APP_DRUPAL_BASE_URL="http://localhost:8888"` instead of `REACT_APP_DRUPAL_BASE_URL="http://127.0.0.1:8888"` you get an annoying error and nothing works

### DIFF
--- a/src/actions/application.js
+++ b/src/actions/application.js
@@ -44,11 +44,12 @@ function* loadMenu() {
   try {
     yield put(resetLoading());
     yield put(showLoading());
-    const menuLinks = yield call(api, 'menu');
+    const maybeMenuLinks = yield call(api, 'menu');
+
     yield put({
       type: MENU_LOADED,
       payload: {
-        menuLinks,
+        maybeMenuLinks,
       },
     });
   } catch (error) {
--- a/src/actions/application.js
+++ b/src/actions/application.js
@@ -35,6 +35,9 @@ function* loadMenu() {
     yield put(resetLoading());
     yield put(showLoading());
     const menuLinks = yield call(api, 'menu');
+    if (menuLinks.message) {
+      throw new Error('Unable to access data from Drupal. Did you set REACT_APP_DRUPAL_BASE_URL to localhost instead of 127.0.0.1?');
+    }
     yield put({
       type: MENU_LOADED,
       payload: {
--- a/src/actions/application.js
+++ b/src/actions/application.js
@@ -36,7 +36,9 @@ function* loadMenu() {
     yield put(showLoading());
     const menuLinks = yield call(api, 'menu');
     if (menuLinks.message) {
-      throw new Error('Unable to access data from Drupal. Did you set REACT_APP_DRUPAL_BASE_URL to localhost instead of 127.0.0.1?');
+      throw new Error(
+        'Unable to access data from Drupal. Did you set REACT_APP_DRUPAL_BASE_URL to localhost instead of 127.0.0.1?',
+      );
     }
     yield put({
       type: MENU_LOADED,
--- a/src/actions/application.js
+++ b/src/actions/application.js
@@ -44,12 +44,12 @@ function* loadMenu() {
   try {
     yield put(resetLoading());
     yield put(showLoading());
-    const maybeMenuLinks = yield call(api, 'menu');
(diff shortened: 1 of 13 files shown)


src/actions/application.js
```diff
@@ -35,6 +35,9 @@ function* loadMenu() {
     yield put(resetLoading());
     yield put(showLoading());
     const menuLinks = yield call(api, 'menu');
+    if (menuLinks.message) {
+      throw new Error('Unable to access data from Drupal. Did you set REACT_APP_DRUPAL_BASE_URL to localhost instead of 127.0.0.1?');
```
dawehner: It is always nice to provide helpful messages like this 👍 Great idea

src/components/06_wrappers/Default/Default.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { func, node, arrayOf, shape, string } from 'prop-types';
+import PropTypes from 'prop-types';
```
dawehner: Seems like an unnecessary changes, but it also seems most people import proptypes like that.

dawehner:
> ci/circleci: lint — Your tests failed on CircleCI
Sadly tests failed...

src/components/06_wrappers/Default/Default.js
justafish: yeah I was trying to figure stuff out with defaultProps to start with and it was really annoying to work with in it's current state

src/actions/application.js
```diff
@@ -35,6 +35,11 @@ function* loadMenu() {
     yield put(resetLoading());
     yield put(showLoading());
     const menuLinks = yield call(api, 'menu');
+    if (menuLinks.message) {
```
dawehner: I'm wondering whether we could look into the cookie and see whether the resulting response had one. ❓

src/reducers/application.js
```diff
@@ -44,7 +44,9 @@ export default (state = initialState, action) => {
     case MENU_LOADED: {
       return {
         ...state,
-        menuLinks: action.payload.menuLinks,
+        menuLinks: Array.isArray(action.payload.menuLinks)
```
dawehner: How about falling back to state.menuLinks?

lauriii:
I'm not sure how to trigger the error message. I tried to point the API to wrong url and I got following error `TypeError: Failed to fetch`.

dawehner:
@lauriii Did you tried using 127.0.0.1 in drupal and localhost in the app configuration? That's how ran into the issue the first time by myself.

src/components/02_atoms/Message/Message.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import { css } from 'emotion';
-import { string, oneOf } from 'prop-types';
```
mattgrill: Are we going to do this all over the place now? Maybe we can make an issue to migrate to this usage of `PropTypes`

src/reducers/application.js
```diff
@@ -38,31 +38,38 @@ export default (state = initialState, action) => {
       };
     }
     case SET_MESSAGE: {
+      // This causes a new messages object to be created, instead of
+      // maintaining a reference to the old data structure.
+      const messages = state.messages.map(message => message);
```
mattgrill: Can we do some `...` operator magic here?

src/reducers/application.js
mattgrill: `action.payload.menuLinks` can we just check the length?

src/components/02_atoms/Message/Message.js
dawehner: >  Maybe we can make an issue to migrate to this usage of PropTypes

dawehner: https://github.com/jsdrupal/drupal-admin-ui/issues/160

src/reducers/application.js
dawehner: Interesting idea!

src/reducers/application.js
justafish: Things which aren't arrays can also have a .length property

src/reducers/application.js
justafish: :+1:

Issue #131, opened 2018-06-11
Create caching strategy for data in store


lauriii:
We might want to consider caching some data in the store instead of refetching everything on every route change. There are some examples such as current username, menu etc. that are quite easy to persist in the store but should be changed when mutations happen. Some more difficult cases are for example content listing and single pieces of content.

Issue #127, opened 2018-06-07
Add bulk operations support


dawehner:
It would be nice to have some level of bulk operations, so you can publish multiple nodes at once.

ol0lll:
#144 fixes this

PR #135, opened 2018-06-13
Simple bulk operation support

dawehner:
This PR adds some bulk operation support, see issue https://github.com/jsdrupal/drupal-admin-ui/issues/127

Right now the UI is really rough, I bet there is a lot which can be done:

* The user selects an action to be executed
* The user selects a bunch of nodes
* "Saving" applies the action on all these nodes
* Right now the endpoints are triggered, but the store itself is untouched.

Related discussion: https://github.com/jsdrupal/drupal-admin-ui/issues/131

TODO:

* [ ] Improve the user experience
* [ ] Figure out how to sync the store with the updates done on the content itself

### DIFF
--- a/admin_ui_support/admin_ui_support.routing.yml
+++ b/admin_ui_support/admin_ui_support.routing.yml
@@ -34,14 +34,14 @@ admin_ui_support.dblog_types:
     _permission: 'access site reports'
     _format: 'json'
 
-#admin_ui_support.dblog.overview:
-#  path: '/vfancy/admin/reports/dblog'
-#  defaults:
-#    _controller: '\Drupal\admin_ui_support\Controller\EmptyController::noop'
-#  options:
-#    _admin_related_route: 'dblog.overview'
-#  requirements:
-#    _permission: 'access site reports'
+admin_ui_support.dblog.overview:
+  path: '/vfancy/admin/reports/dblog'
+  defaults:
+    _controller: '\Drupal\admin_ui_support\Controller\EmptyController::noop'
+  options:
+    _admin_related_route: 'dblog.overview'
+  requirements:
+    _permission: 'access site reports'
 
 admin_ui_support.admin_ui_support_settings:
   path: '/admin/config/system/admin-ui-support'
--- a/admin_ui_support/admin_ui_support.routing.yml
+++ b/admin_ui_support/admin_ui_support.routing.yml
@@ -34,14 +34,14 @@ admin_ui_support.dblog_types:
     _permission: 'access site reports'
     _format: 'json'
 
-admin_ui_support.dblog.overview:
-  path: '/vfancy/admin/reports/dblog'
-  defaults:
-    _controller: '\Drupal\admin_ui_support\Controller\EmptyController::noop'
-  options:
-    _admin_related_route: 'dblog.overview'
-  requirements:
-    _permission: 'access site reports'
+#admin_ui_support.dblog.overview:
+#  path: '/vfancy/admin/reports/dblog'
+#  defaults:
+#    _controller: '\Drupal\admin_ui_support\Controller\EmptyController::noop'
+#  options:
+#    _admin_related_route: 'dblog.overview'
+#  requirements:
+#    _permission: 'access site reports'
 
 admin_ui_support.admin_ui_support_settings:
   path: '/admin/config/system/admin-ui-support'
--- a/src/actions/application.js
+++ b/src/actions/application.js
@@ -103,8 +103,6 @@ function* loadActions() {
     });
   } catch (error) {
     yield put(setMessage(error.toString()));
-  } finally {
-    yield put(hideLoading());
   }
 }
 
--- a/src/actions/application.js
+++ b/src/actions/application.js
@@ -78,7 +78,38 @@ function* loadContentTypes() {
   }
 }
 
+/**
+ * Gets all available action types.
+ */
+export const ACTIONS_REQUESTED = 'ACTIONS_REQUESTED';
+export const requestActions = () => ({
+  type: ACTIONS_REQUESTED,
+  payload: {},
+});
+
+export const getActionsCache = state => state.application.actions;
+export const ACTIONS_LOADED = 'ACTIONS_LOADED';
+function* loadActions() {
+  try {
+    let actions = yield select(getActionsCache);
+    if (!Object.keys(actions).length) {
+      actions = yield call(api, 'actions');
+    }
+    yield put({
+      type: ACTIONS_LOADED,
+      payload: {
+        actions,
+      },
+    });
+  } catch (error) {
+    yield put(setMessage(error.toString()));
+  } finally {
+    yield put(hideLoading());
+  }
+}
+
 export default function* watchRequestedMenu() {
   yield takeLatest(MENU_REQUESTED, loadMenu);
   yield takeLatest(CONTENT_TYPES_REQUESTED, loadContentTypes);
+  yield takeLatest(ACTIONS_REQUESTED, loadActions);
 }
--- a/src/actions/content.js
+++ b/src/actions/content.js
@@ -119,13 +119,15 @@ export const KNOWN_ACTIONS = [
 ];
 
 // @todo Find a better name
+// @todo How do we update the store with the new values of the nodes
+//    or the deleted nodes.
 export function* doActionExecute({ payload: { action, nids } }) {
   try {
     const contentList = yield select(state => state.content.contentList);
     const actions = nids
       .map(nid => {
         const node = contentList.filter(
-          contentItem => contentItem.attributes.nid == nid,
+          contentItem => String(contentItem.attributes.nid) === nid,
         )[0];
 
         let saveAction;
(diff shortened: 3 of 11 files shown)


src/actions/application.js
```diff
@@ -78,7 +78,38 @@ function* loadContentTypes() {
   }
 }
 
+/**
+ * Gets all available action types.
+ */
+export const ACTIONS_REQUESTED = 'ACTIONS_REQUESTED';
```
dawehner: I considered the list of available actions to be some form of global information. Not sure though how useful it is to share this.

src/actions/content.js
```diff
@@ -97,7 +97,143 @@ function* loadContent(action) {
...
+  try {
+    const contentList = yield select(state => state.content.contentList);
+    const actions = nids
+      .map(nid => {
+        const node = contentList.filter(
+          contentItem => String(contentItem.attributes.nid) === nid,
+        )[0];
+
+        let saveAction;
+        switch (action.attributes.plugin) {
```
dawehner: Instead of doing the API calls directly I'm wondering whether this should trigger NODE_DELETE and NODE_SAVE actions, which then update the store, but also trigger the API calls.❓

src/components/05_pages/Content/Content.js
```diff
@@ -175,9 +195,38 @@ class Content extends Component {
...
+                </MenuItem>
+                {this.props.actions.map(action => (
+                  <MenuItem key={action.id} value={action.attributes.id}>
+                    {action.attributes.label}
+                  </MenuItem>
+                ))}
+              </Select>
+            </FormControl>
+          )}
+          {this.state.action && (
```
dawehner: Is there a better word than "Save"? Maybe apply

src/components/06_wrappers/Default/Default.js
```diff
@@ -113,6 +113,9 @@ const mapStateToProps = state => ({
   menuLinks: state.application.menuLinks,
 });
 
-export default connect(mapStateToProps, {
-  requestMenu,
-})(Default);
+export default connect(
```
dawehner: It is interesting that prettier triggered this suddenly.

src/utils/api/api.js
```diff
@@ -40,10 +44,50 @@ async function api(
...
     case 'contentTypes':
       url = '/jsonapi/node_type';
       options.headers.Accept = 'application/vnd.api+json';
       break;
+    case 'node:delete':
+      // Set the type to the right value for jsonapi to process.
+      // @todo Ideally this should not be differnet in the first place.
+      parameters.node = {
+        ...parameters.node,
+        type: parameters.node.type.includes('--') ? parameters.node.type : `node--${parameters.node.type}`,
```
dawehner: @lauriii This was the issue I had which caused the 500

src/actions/content.js
```diff
@@ -97,7 +97,143 @@ function* loadContent(action) {
...
+  return {
+    type: ACTION_EXECUTE,
+    payload: {
+      action,
+      nids,
+    },
+  };
+};
+
+export const KNOWN_ACTIONS = [
```
lauriii: Would it be more accurate to call this `SUPPORTED_ACTIONS`?

admin_ui_support/admin_ui_support.routing.yml
```diff
@@ -34,14 +34,14 @@ admin_ui_support.dblog_types:
...
 
-admin_ui_support.dblog.overview:
-  path: '/vfancy/admin/reports/dblog'
-  defaults:
-    _controller: '\Drupal\admin_ui_support\Controller\EmptyController::noop'
-  options:
-    _admin_related_route: 'dblog.overview'
-  requirements:
-    _permission: 'access site reports'
+#admin_ui_support.dblog.overview:
```
lauriii: We should probably just delete these for now. We can use git to go back anytime if we think it's valuable.

src/components/05_pages/Content/Content.js
lauriii: I think `Apply` is definitely better than `Save`. Other options that pop into my mind would be `Execute` or just `Submit`.

src/components/05_pages/Content/Content.js
```diff
@@ -76,22 +83,35 @@ class Content extends Component {
...
   static defaultProps = {
     contentList: [],
     includes: {},
+    actions: [],
   };
   state = {
     contentTypes: [],
     status: null,
+    action: null,
+    checked: {},
```
lauriii: What do you think should happen if user has selected checkboxes and uses pagination to navigate to the next page? The current Drupal UI discards the checked items from previous pages but with the architecture of this application we could still take them into account. 🤔

src/components/05_pages/Content/Content.js
```diff
@@ -175,9 +195,38 @@ class Content extends Component {
           </Button>
         </div>
 
+        <div className={styles.filters}>
+          {this.props.actions && (
+            <FormControl className={styles.formControl}>
```
lauriii: This class adds some margin to the element which makes it look a bit strange.
screen shot 2018-06-14 at 15 09 34

src/utils/api/api.js
```diff
@@ -55,7 +99,12 @@ async function api(
         : ''
     }`,
     options,
-  ).then(res => res.json());
+  ).then(res => {
+    if (options.text) {
+      return res.text();
```
lauriii: Why do we need this? Shouldn't the backend always return json?

src/components/05_pages/Content/Content.js
dawehner: Thank you! I went with Apply, but ultimately the interaction model for VBO might change quite a bit anyway.

src/actions/content.js
dawehner: Good idea 👍

admin_ui_support/admin_ui_support.routing.yml
dawehner: Yeah let me remove this change from the PR. I needed the page to be able to see the 500s properly.

src/components/05_pages/Content/Content.js
dawehner: mhhh, there is something in me which would like to introduce a confirm page/model/something for every action, not just dangerous ones like unpublish. When we have this confirm page, keeping the state across multiple pages seems more safe.

src/components/05_pages/Content/Content.js
dawehner: I fixed it, 
screen shot 2018-06-15 at 17 47 56

src/utils/api/api.js
dawehner: I added a comment. This is for CSRF token support.

src/actions/content.js
dawehner: I implemented this

src/actions/content.js
```diff
@@ -100,6 +100,178 @@ function* loadContent(action) {
...
+  'node_make_sticky_action',
+  'node_make_unsticky_action',
+  'node_promote_action',
+  'entity:publish_action:node',
+  'node_unpromote_action',
+  'entity:unpublish_action:node',
+];
+
+// @todo Find a better name
+// @todo How do we update the store with the new values of the nodes
```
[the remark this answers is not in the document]
dawehner: +:1+

src/actions/content.js
```diff
@@ -100,6 +100,178 @@ function* loadContent(action) {
...
+  'entity:delete_action:node',
+  'node_make_sticky_action',
+  'node_make_unsticky_action',
+  'node_promote_action',
+  'entity:publish_action:node',
+  'node_unpromote_action',
+  'entity:unpublish_action:node',
+];
+
+// @todo Find a better name
```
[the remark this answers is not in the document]
dawehner: Sure, let's do that. It feels weird to have it the other way round compared to the content, but we can always refactor :)

src/actions/content.js
```diff
@@ -100,6 +100,178 @@ function* loadContent(action) {
...
+// @todo How do we update the store with the new values of the nodes
+//    or the deleted nodes.
+export function* doActionExecute({ payload: { action, nids } }) {
```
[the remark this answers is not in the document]
dawehner: Sorry I'm not sure what you mean with that. Do you suggest we should have some form of VBO plugin system in JS?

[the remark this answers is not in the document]
lauriii: I talked with Wim and Gabe and they mentioned that JSON API will support bulk operations in future. I assume it means instead of sending separate HTTP requests for each action, we would be sending multiple requests in a single HTTP request.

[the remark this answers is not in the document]
dawehner: Oh that makes sense! Sure, let's add a quick todo

martinfrances107:
This looks good to me. it functions as expected -- which really good...

I can see one small flaw. 

When I look in chrome devtools at the console output when building the page for the first time 
I see this error.

> index.js:2178 Warning: Failed prop type: Invalid prop `variant` of value `contained` supplied to `Button`, expected one of ["flat","outlined","raised","fab"]
 
in short "contained" is not a valid property value .. I hope this helps.

+              <Button
+                onClick={this.executeAction}
+                color="primary"
+                variant="contained"
+              >

lauriii:
#144 solves this! Thank you 🙂

lauriii:
@martinfrances107 thanks for the feedback! I forgot to mention that your feedback was also addressed as part of #144 🙂

dawehner:
Thank you for finishing this up!

PR #137, opened 2018-06-21
Content list pagination

mattgrill:


src/components/05_pages/Content/Content.js
```diff
@@ -286,6 +310,12 @@ class Content extends Component {
             )}
           </TableBody>
         </Table>
+        <Button variant="flat" onClick={this.previousHandler}>
```
lauriii: We should use the API from Material UI for the pagination links https://material-ui.com/demos/tables/#sorting-amp-selecting

src/components/05_pages/Content/Content.js
```diff
@@ -89,6 +89,10 @@ class Content extends Component {
     contentTypes: [],
     status: null,
     sort: { path: 'changed', direction: 'DESC' },
+    page: {
+      offset: 0,
+      limit: 5,
```
lauriii: Should we show 50 items per page which is default in the current content listing?

src/components/05_pages/Content/Content.js
mattgrill: _fwiw_, `<TablePagination />` looks like it has a bunch of extra props that we can't really provide.

mattgrill:
react_app

@lauriii Looks like we might need some minor styling adjustments.

src/actions/content.js
```diff
@@ -34,6 +35,11 @@ function* loadContent(action) {
       filter: {},
     };
 
+    if (page) {
+      const { offset, limit } = page;
+      queryString.page = { offset, limit };
```
dawehner: Is there a reason we don't just do ```queryString.page = page;```?

src/components/05_pages/Content/Content.js
```diff
@@ -80,221 +80,286 @@ class Content extends Component {
...
+    }
+  }
   tableSortHandler = (path, direction) => () => {
-    this.setState({ sort: { path, direction } }, () => {
-      this.props.requestContent(this.state);
-    });
+    this.setState(
+      {
+        sort: { path, direction },
+        page: { offset: 0, limit: this.state.page.limit },
```
dawehner: I guess it is fine to jump to the first page when sorting. For now at least it is totally fine, and its rather a UX team decision how it should behave.

src/components/05_pages/Content/Content.js
```diff
@@ -80,221 +80,286 @@ class Content extends Component {
...
+              </TableBody>
+            </Table>
+            <TablePagination
+              component="div"
+              count={count}
+              rowsPerPage={limit}
+              page={offset / limit}
+              onChangePage={this.pageChangeHandler}
+              rowsPerPageOptions={[limit]}
+              labelDisplayedRows={() => {}}
```
dawehner: What is the reason we don't show the current page we are on? i think this would be beneficial for the user. ❓

src/actions/content.js
lauriii: I saw this and was thinking the same. I assumed the reason was that if we extend our local state, we might very easily leak that information in the HTTP requests because it would be easy to forget that the whole page object is added to the URL.

src/components/05_pages/Content/Content.js
lauriii: Good suggestion! Done ✅

src/components/05_pages/Content/Content.js
```diff
@@ -80,221 +80,286 @@ class Content extends Component {
...
+    },
   };
   componentDidMount() {
     this.props.requestContentTypes();
     this.props.requestContent(this.state);
   }
+  componentDidUpdate(prevProps) {
+    if (
+      this.state.scrollToTop &&
+      JSON.stringify(prevProps.contentList) !==
```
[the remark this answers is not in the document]
lauriii: Added some documentation here. Is it clearer now?

src/actions/content.js
dawehner: I cannot imagine of usecase to put more state into ```state.page```, but also ¯\_(ツ)_/¯

src/components/05_pages/Content/Content.js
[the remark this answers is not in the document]
dawehner: It is, thank you

PR #140, opened 2018-06-25
Allow closing drawer

lauriii:
screen shot 2018-06-25 at 16 29 23


src/components/06_wrappers/Default/Default.js
```diff
@@ -38,14 +41,50 @@ const iconMap = {
 };
 
 class Default extends React.Component {
+  state = {
+    drawerOpen: false,
```
mattgrill: If you transition between pages does the state of the drawer remain the same? Is this something that should be better kept within Redux?

lauriii: Good point!

Issue #52, opened 2018-03-26
Add nightwatch for functional testing


justafish:
This will be a separate set of tests from the ones that will be available in [Drupal core](https://www.drupal.org/project/drupal/issues/2869825), although it would be nice if the folder structure was the same so that core would also pick up these tests.

PR #141, opened 2018-06-25
[WIP][#52] Add Nightwatch for functional browser tests, plus a test for th…

justafish:
…e Menu
#52 
Todo:
- [ ] Documentation
- [ ] Get this running in CircleCI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,57 +2,112 @@ aliases:
   - &default
     docker:
       - image: circleci/node:8-browsers
+
+  - &node-modules-key
+    v3-yarn-dependencies-{{ checksum "yarn.lock" }}
+
+  - &build-key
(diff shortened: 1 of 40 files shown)


dawehner:
@justafish Do you plan to hook this up with circleCI?

justafish:
@dawehner I do, but if someone else wants to give it a go before I get to it then ping me in #javascript on Drupal Slack :+1:

dawehner:
I do have one question right now though. This tests the app outside of Drupal. I am wondering whether that's the right thing to do. Should we maybe test the app being part of Drupal, aka. symlinked in? This way we might catch some bugs we otherwise would miss?

Once we say that we should test the app inside Drupal, couldn't we use the existing nightwatch setup?

justafish:
Closed for #170

Issue #119, opened 2018-06-06
Allow deleting content inside the app


lauriii:
Deleting content works currently by redirecting to Drupal's delete content form. We could improve the UX by allowing deleting content directly on the app. Material UI has tools in place for this use case: https://material.io/design/components/dialogs.html#alert-dialog

iberdinsky-skilld:
One thing not obvious here. According to React practice we have to create new dialog component. But i think it should be in atoms and reusable by other app components. and in this case it should be component: Dialog with alert variant here.

martinfrances107:
@mattgrill  -- Are you actively working on this  @?... I have a first draft I am about to create a pull request

I am just posting a early draft - so we can grasp the rough shape of this.

I just copied the example source code from the alert dialog button here 
https://material-ui.com/demos/dialogs/#alerts

For D8 purposes DeleteButton may not be generic enough -- maybe a first cut will highlight any problem areas.

It is not functional ... it just displays a modal which matches the text produced by D8.6's confirm form and then does NOTHING

MY TODO list

1) There are some styling snags .. the shape of the new delete button needs to be circular like the other buttons 
2) Make pressing confirm on the dialog functional.

PS thanks for taking the time to describe an novice issue lauriii++

lauriii:
@martinfrances107 I posted feedback on the UI at the PR! I haven't heard @mattgrill mentioning this issue for a while so I'm unassigning him from here for now 🙂

dawehner:
@martinfrances107 Some of the changes in https://github.com/jsdrupal/drupal-admin-ui/pull/135 could be helpful for you too. It for example provides delete support on the API level and provides a redux action as well.

martinfrances107:
Here are my notes associated with the next PR.

@lauriii  Thanks for the feedback ... after reading your notes and after a little thinking,  I have made the button more component like.

 ... As a component I want a generic  button that lives in a Drupal 'operations' column for whatever table and function comes our way next. One that can be labelled as 'add' 'edit' whatever..and can be styled with a Icon image as per the want of the next developer.

The name I went with was "OpsModalButton" - but I don't particularly feel married to it.

On my todo list -- I still have to fix up your feedback about the in modal button labels.

But in the next iteration I am going to work on the more meaty side of the problem -- and use dawehner's suggestion to implement  the required function. ( Thanks dawehner BTW)

martinfrances107:
The missing work .....

[...]

PR #142, opened 2018-06-26
First cut - Allow deleting content inside the app #119

martinfrances107:
This is a first cut 

I have described what is going on in this comment 

https://github.com/jsdrupal/drupal-admin-ui/issues/119#issuecomment-400264640

### DIFF
--- a/src/components/02_atoms/DeleteButton/DeleteButton.js
+++ b/src/components/02_atoms/DeleteButton/DeleteButton.js
@@ -1,0 +1,70 @@
+import React from 'react';
+import Button from '@material-ui/core/Button';
+import DeleteIcon from '@material-ui/icons/Delete';
+import Dialog from '@material-ui/core/Dialog';
+import DialogActions from '@material-ui/core/DialogActions';
+import DialogContent from '@material-ui/core/DialogContent';
+import DialogContentText from '@material-ui/core/DialogContentText';
+import DialogTitle from '@material-ui/core/DialogTitle';
+import Slide from '@material-ui/core/Slide';
+
+import PropTypes from 'prop-types';
+
+function Transition(props) {
+  return <Slide direction="up" {...props} />;
+}
+
+class AlertDialogSlide extends React.Component {
+  state = {
+    open: false,
+  };
+
+  handleClickOpen = () => {
+    this.setState({ open: true });
+  };
+
+  handleClose = () => {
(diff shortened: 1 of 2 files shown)


src/components/02_atoms/DeleteButton/DeleteButton.js
```diff
@@ -0,0 +1,70 @@
...
+            <DialogContentText id="alert-dialog-slide-description">
+              {this.props.text}
+            </DialogContentText>
+          </DialogContent>
+          <DialogActions>
+            <Button onClick={this.handleClose} color="primary">
+              Disagree
+            </Button>
+            <Button onClick={this.handleClose} color="primary">
+              Agree
```
lauriii: We could probably use the texts from the buttons in the current UI since I assume they are user tested. They are currently "Delete" and "Cancel".

src/components/02_atoms/DeleteButton/DeleteButton.js
```diff
@@ -0,0 +1,70 @@
...
+          <DialogContent>
+            <DialogContentText id="alert-dialog-slide-description">
+              {this.props.text}
+            </DialogContentText>
+          </DialogContent>
+          <DialogActions>
+            <Button onClick={this.handleClose} color="primary">
+              Disagree
+            </Button>
+            <Button onClick={this.handleClose} color="primary">
```
lauriii: Wondering if we should follow the pattern from current Drupal UI where the buttons have different colors. I'm not keen to the current solution but just wanted to raise it up to make sure we at least give some thought for it.

src/components/05_pages/Content/Content.js
```diff
@@ -272,14 +272,10 @@ class Content extends Component {
...
                     </IconButton>
-                    <IconButton
-                      aria-label="delete"
-                      className={styles.button}
-                      component={Link}
-                      to={`/node/${nid}/delete`}
-                    >
-                      <DeleteIcon />
-                    </IconButton>
+                    <DeleteButton
```
lauriii: We should make sure that current styles still apply for the new icon. It seems like the positioning and the color is a bit off.

screen shot 2018-06-26 at 18 01 17

martinfrances107:
I am still learning good github flow and how to collaborate... sorry for the noise

I will push all future updates to 

 #143

PR #143, opened 2018-06-26
More like a component - Allow deleting content inside the app #119

martinfrances107:
Making the new button more reusable.

### DIFF
--- a/src/components/02_atoms/OpsModalButton/OpsModalButton.js
+++ b/src/components/02_atoms/OpsModalButton/OpsModalButton.js
@@ -1,0 +1,83 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Button from '@material-ui/core/Button';
+import Dialog from '@material-ui/core/Dialog';
+import DialogActions from '@material-ui/core/DialogActions';
+import DialogContent from '@material-ui/core/DialogContent';
+import DialogContentText from '@material-ui/core/DialogContentText';
+import DialogTitle from '@material-ui/core/DialogTitle';
+import IconButton from '@material-ui/core/IconButton';
+import Slide from '@material-ui/core/Slide';
+
+function Transition(props) {
+  return <Slide direction="up" {...props} />;
+}
+
+/* A Drupal specific button, for use in the operations column of a table.
+ *
+ * Typically a Add, Edit or Delete icon can be provided as a child element
+ * see for example  '@material-ui/icons/Add'
+ *
+ * When pressed a modal dialog slides up into view.
+ *
+ * Custom Properties :-
+ *   title, string to appear in the modal dialog.
+ *   text, string to appear below the title.
+ */
+class OpsModalButton extends React.Component {
+  state = {
+    open: false,
+  };
+
+  handleClickOpen = () => {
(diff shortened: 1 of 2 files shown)


src/components/02_atoms/OpsModalButton/OpsModalButton.js
```diff
@@ -0,0 +1,83 @@
...
+
+    return (
+      <IconButton {...buttonProps} onClick={this.handleClickOpen}>
+        {this.props.children}
+        <Dialog
+          open={this.state.open}
+          TransitionComponent={Transition}
+          keepMounted="keepMounted"
+          onClose={this.handleClose}
+          aria-labelledby="alert-dialog-slide-title"
```
dawehner: I am wondering what happens if something reuses this component? Wouldn't this lead to duplicate IDs? One for each row for example? Maybe we need to allow to pass in a prefix? ❓

src/components/02_atoms/OpsModalButton/OpsModalButton.js
```diff
@@ -0,0 +1,83 @@
...
+        </Dialog>
+      </IconButton>
+    );
+  }
+}
+
+OpsModalButton.propTypes = {
+  title: PropTypes.string.isRequired,
+  text: PropTypes.string.isRequired,
+  children: PropTypes.oneOfType([
```
dawehner: 🔧 As you can see on https://github.com/facebook/prop-types/blob/master/factoryWithTypeCheckers.js#L435 it is enough to use ```PropTypes.node```, arrays of nodes are covered.

dawehner:
@martinfrances107 does this mean https://github.com/jsdrupal/drupal-admin-ui/pull/142 can be closed?

src/components/02_atoms/OpsModalButton/OpsModalButton.js
martinfrances107: Thanks for the tip.... I will work this information into the next iteration.

martinfrances107:
@dawehner when you say 

> does this mean #142 can be closed?

Sorry I haven't figured out the correct github workflow yet... I will try to stick to just one PR in the future. Yes I think 142 can be closed.

src/components/02_atoms/OpsModalButton/OpsModalButton.js
martinfrances107: Thanks for highlighting potential problems. It is really helpful -- that section of the code just some sample code.... copied from the material-ui website so I may have to component-ize it.

It is on my todo list ... that is I don't know the correct response yet... but I will work out an informed response by the next iteration.

martinfrances107:
A) Reroll 

B) Regarding the  confirm modal's button labels ... "Delete" and "Cancel" 
They are now consistent with confirm form it is replacing.

C) I have simplified the definition of Simplified OpsModalButton:PropType

I am going to shift my focus... I think this should be postponed until #135 is complete as this will provide the delete action.

lauriii:
Thanks for the PR! We used this as a base with @ol0lll at the Drupal Dev Days sprints for creating #145.

PR #146, opened 2018-07-03
Followup to #145 document properties of the new button.

martinfrances107:
At the last moment when  #145 was being stitched together with the bulk operations issue the button labels where added as additional properties to the opModalButton.. this is a good thing to do 

But the annotation at the start of the class definition was not updated.

This is just a minor tidyup

### DIFF
--- a/src/components/02_atoms/OpsModalButton/OpsModalButton.js
+++ b/src/components/02_atoms/OpsModalButton/OpsModalButton.js
@@ -23,6 +23,9 @@ function Transition(props) {
  * Custom Properties :-
  *   title, string to appear in the modal dialog.
  *   text, string to appear below the title.
+ *   cancelText, string to appear in the cancel button.
+ *   confirmText, string to appear in the confirm button.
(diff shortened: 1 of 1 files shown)


src/components/02_atoms/OpsModalButton/OpsModalButton.js
```diff
@@ -23,6 +23,9 @@ function Transition(props) {
  * Custom Properties :-
  *   title, string to appear in the modal dialog.
  *   text, string to appear below the title.
+ *   cancelText, string to appear in the cancel button.
```
dawehner: I an wondering: Could we document proptypes like this? https://react-styleguidist.js.org/docs/documenting.html This seems to be a standard which we could show in styleguides at some point.

martinfrances107: Thanks for the feedback ... a good idea as usual, I was kinda making it up as I went along.  I have updated the PR. Also this is the first time we are using the "extends React.component" template ... I looked for other places in the code base to also modify.. but hey nothing doing

PR #148, opened 2018-07-04
Initial node form, introducting of ui metadata and widgets mapping

dawehner:
After talking with @lauriii @justafish @mattgrill yesterday and outlined in https://docs.google.com/document/d/1QKTPgSnJCk94Pbu_mIJyIh-99fly4c0N5FEBHcwwyBw/edit
here is an initial start

### DIFF
--- a/src/components/05_pages/NodeForm/UiMetadata.js
+++ b/src/components/05_pages/NodeForm/UiMetadata.js
@@ -1,0 +1,8 @@
+const UiMetadata = {
+  title: {
+    widget: 'string_textfield',
+    constraints: [],
+  },
+};
+
+export default UiMetadata;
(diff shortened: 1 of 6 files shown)


src/components/05_pages/NodeForm/UiMetadata.js
```diff
@@ -0,0 +1,8 @@
+const UiMetadata = {
```
dawehner: ℹ️This would be fetched from the API.

src/components/05_pages/NodeForm/Widgets.js
```diff
@@ -0,0 +1,7 @@
+import StringTextfield from '../../02_atoms/Widgets/StringTextfield';
+
+const widgets = {
+  string_textfield: StringTextfield,
```
dawehner: For now this is hard coded but I could imagine that this will be part of the build step and generated at some point.

Issue #152, opened 2018-07-04
Add support for adding new content with the 'plus' button.


mattgrill:
The plus button currently does not offer the same functionality that `/node/add` normally would.

- all content types are currently loaded for the content listing page. 

<img width="221" alt="react_app" src="https://user-images.githubusercontent.com/1090713/42275683-090c7ad2-7f89-11e8-9b04-131de2e5daca.png">

lauriii:
This is fixed by #159

PR #159, opened 2018-07-04
Implement add content page

ol0lll:
Implement #152 
Todo:
~~* 'node/add' links~~
~~* path override in admin_ui_support~~

### DIFF
--- a/admin_ui_support/admin_ui_support.routing.yml
+++ b/admin_ui_support/admin_ui_support.routing.yml
@@ -61,3 +61,13 @@ admin_ui_support.system.admin_content:
     _admin_related_route: 'system.admin_content'
   requirements:
     _permission: 'access administration pages'
+
+admin_ui_support.node.add_page:
+  path: '/vfancy/node/add'
+  defaults:
+    _controller: '\Drupal\admin_ui_support\Controller\EmptyController::noop'
+  options:
+    _admin_related_route: 'node.add_page'
+    _node_operation_route: TRUE
+  requirements:
+    _node_add_access: 'node'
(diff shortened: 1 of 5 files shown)


src/components/05_pages/AddContent/AddContent.js
```diff
@@ -0,0 +1,58 @@
...
+    requestContentTypes: PropTypes.func.isRequired,
+  };
+  componentDidMount() {
+    this.props.requestContentTypes();
+  }
+  render = () => {
+    return (
+      <div className={styles.root}>
+        <Paper>
+          {console.log(this.props.contentTypes)}
```
dawehner: I guess we could drop that :)

src/components/05_pages/AddContent/AddContent.js
```diff
@@ -0,0 +1,58 @@
...
+          <List>
+            {Object.keys(this.props.contentTypes).map(contentType => (
+              <ListItem key={contentType}>
+                <ListItemText
+                  primary={this.props.contentTypes[contentType].name}
+                  secondary={(<div dangerouslySetInnerHTML={{ __html: this.props.contentTypes[contentType].description}} />)}
```
dawehner: 🔧 Note: We do have the ```interweave``` module which strips dangerous HTML from the description

ol0lll:
I moved the page route to _content/add_ because I could not figure out how to make subpaths like _node/add/<content_type>_ work when the app is overriding _node/add_. Do we need an issue for that?

dawehner:
How about applying this PR on top of it: https://github.com/ol0lll/drupal-admin-ui/pull/1

admin_ui_support/admin_ui_support.routing.yml
```diff
@@ -61,3 +61,13 @@ admin_ui_support.system.admin_content:
     _admin_related_route: 'system.admin_content'
   requirements:
     _permission: 'access administration pages'
+
+admin_ui_support.node.add_page:
```
dawehner: Thank you for adding it!

Issue #160, opened 2018-07-05
Import Proptype from 'prop-types' instead of destructing


dawehner:
See conversation at https://github.com/jsdrupal/drupal-admin-ui/pull/134#discussion_r200188457

* Convert all the proptypes to use ```Proptype.string``` etc.

rsanzante:
I understand is something like this: https://github.com/rsanzante/drupal-admin-ui/commit/e61cbaf172b6eb1d1308fcecbaf231a1e68e18be

dawehner:
Yes, exactly!

PR #161, opened 2018-07-05
Import proptypes object

rsanzante:
Solves https://github.com/jsdrupal/drupal-admin-ui/issues/160.

### DIFF
--- a/src/InitialRedirect.js
+++ b/src/InitialRedirect.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { shape, string } from 'prop-types';
+import PropTypes from 'prop-types';
 import { Redirect } from 'react-router-dom';
 
 const InitialRedirect = ({ location: { search } }) => {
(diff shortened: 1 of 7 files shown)


dawehner:
@rsanzante Do you mind running prettier once :)

rsanzante:
Oops, some are coding standard but the las two errors seem not declared prop validations. Not good to fix two things in the same PR but I guess it's ok for this case.

```
drupal-admin-ui-demo/drupal-admin-ui-demo/drupal-admin-ui/src/components/05_pages/Reports/Dblog.js
   84:41  error  'filterOptions.offset' is missing in props validation  react/prop-types
  104:46  error  'filterOptions.types' is missing in props validation   react/prop-types

```

rsanzante:
I didn't see your comment but I managed to find how to run prettier, I wasn't aware it was available, thanks!

Issue #156, opened 2018-07-04
Checkbox field widget


dawehner:
For the "Published" checkbox we need a checkbox field widget, see ```core/lib/Drupal/Core/Field/Plugin/Field/FieldWidget/BooleanCheckboxWidget.php:20``` and https://github.com/jsdrupal/drupal-admin-ui/pull/148/files#diff-4ed0700d6e4aa3489751ab950e7bbe5a

VD39:
I will work on this issue.

rsanzante:
Sorry, I didn't post I was facing it.

VD39:
As this has been merged, I assume this issue can now be closed?

dawehner:
Totally!

PR #166, opened 2018-07-05
Boolean checkbox widget

rsanzante:
Shaped from current StringTextfield widget.

https://github.com/jsdrupal/drupal-admin-ui/issues/156

Not really sure this what is wanted.

### DIFF
--- a/src/components/02_atoms/Widgets/BooleanCheckbox.js
+++ b/src/components/02_atoms/Widgets/BooleanCheckbox.js
@@ -13,7 +13,6 @@ const BooleanCheckbox = props => {
       control={
         <CheckBox
           id={`${props.fieldName}-cb`}
-          checked={value}
           onChange={event => onChange(event.target.value)}
           margin="normal"
           value={value}
--- a/src/components/02_atoms/Widgets/BooleanCheckbox.js
+++ b/src/components/02_atoms/Widgets/BooleanCheckbox.js
@@ -15,7 +15,7 @@ const BooleanCheckbox = props => (
 
 BooleanCheckbox.propTypes = {
   ...WidgetPropTypes,
-  checked: PropTypes.boolean,
+  checked: PropTypes.bool,
   label: PropTypes.string,
 };
 
--- a/src/components/02_atoms/Widgets/BooleanCheckbox.js
+++ b/src/components/02_atoms/Widgets/BooleanCheckbox.js
@@ -1,27 +1,37 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import CheckBox from '@material-ui/core/Checkbox';
+import FormControlLabel from '@material-ui/core/FormControlLabel';
 import WidgetPropTypes from '../../05_pages/NodeForm/WidgetPropTypes';
 
-const BooleanCheckbox = props => (
-  <CheckBox
-    id={props.fieldName}
-    checked={props.checked}
-    onChange={event => props.onChange(event.target.value)}
-    margin="normal"
-    label={props.label}
-  />
-);
+const BooleanCheckbox = props => {
+  const { onChange, label, id, value, ...checkboxProps } = props;
+
+  return (
+    <FormControlLabel
+      id={`${props.fieldName}-label`}
+      control={
+        <CheckBox
+          id={`${props.fieldName}-cb`}
+          checked={value}
+          onChange={event => onChange(event.target.value)}
+          margin="normal"
+          value={value}
+          {...checkboxProps}
+        />
+      }
+      label={label}
+    />
+  );
+};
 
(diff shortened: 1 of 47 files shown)


rsanzante:
Thanks for your review! I think I've included it.

rsanzante:
Oops, conflict resolved but now eslint complains about prettier errors. FIxing them (although I think I got them during master merge to resolve conflict).

rsanzante:
Solved. I wonder if eslint properly configured because some changes seem ugly to me. It forced me to put multiline declaration into one line harder to read. See my last commit.

dawehner:
It doesn't work properly right now. We need something like this at least I believe

```
diff --git a/src/components/02_atoms/Widgets/BooleanCheckbox.js b/src/components/02_atoms/Widgets/BooleanCheckbox.js
index a5bd85e..5fec7c3 100644
--- a/src/components/02_atoms/Widgets/BooleanCheckbox.js
+++ b/src/components/02_atoms/Widgets/BooleanCheckbox.js
@@ -6,8 +6,8 @@ import WidgetPropTypes from '../../05_pages/NodeForm/WidgetPropTypes';
 const BooleanCheckbox = props => (
   <CheckBox
     id={props.fieldName}
-    checked={props.checked}
-    onChange={event => props.onChange(event.target.value)}
+    checked={props.value}
+    onChange={event => props.onChange(event.target.checked)}
     margin="normal"
     label={props.label}
   />
@@ -15,12 +15,12 @@ const BooleanCheckbox = props => (
 
 BooleanCheckbox.propTypes = {
   ...WidgetPropTypes,
-  checked: PropTypes.bool,
+  value: PropTypes.bool,
   label: PropTypes.string,
 };
 
 BooleanCheckbox.defaultProps = {
-  checked: false,
+  value: false,
   label: 'On',
 };
 

```

src/components/02_atoms/Widgets/BooleanCheckbox.js
```diff
@@ -0,0 +1,27 @@
...
+import CheckBox from '@material-ui/core/Checkbox';
+import WidgetPropTypes from '../../05_pages/NodeForm/WidgetPropTypes';
+
+const BooleanCheckbox = props => (
+  <CheckBox
+    id={props.fieldName}
+    checked={props.checked}
+    onChange={event => props.onChange(event.target.value)}
+    margin="normal"
+    label={props.label}
```
martinfrances107: label is not a valid property here 

can I suggest wrapping checkbox in FormControlLabel
I am also thinking about how to make this extendable and want to pass any extra properties like color="primary" down into the checkbox element.

I hope this helps

  render() {
    /* Pass all props not used at this level down into Checkbox */
    const { onChange, label, ...checkboxProps } = this.props;
    return (
      <FormControlLabel
        control={
          <Checkbox
            {...checkboxProps}
            checked={this.state.checked}
            onChange={onChange}
            value=""
          />
        }
        label={label}
      />
    );
  }

dawehner: Great remark @martinfrances107 !

rsanzante:
Fixes:
  - Wrap CheckBox with a FormControlLabel.
  - Pass addtional props to CheckBox.
  - Fix value/checked prop.
  - Added id to both elements using fieldName and a suffix (-label and -cb).

francescopersico:
Yes, what i mean is that the only files included in this PR should be:

BooleanCheckbox.js
UiMetadata.js
Widgets.js

Other files are not related to the PR, and if they need formatting/linting fixes should be done outside this PR.

dawehner:
@francescopersico Oops I think I misread your message on my phone :)

src/components/02_atoms/Widgets/BooleanCheckbox.js
```diff
@@ -0,0 +1,36 @@
...
+const BooleanCheckbox = props => {
+  const { onChange, label, id, value, ...checkboxProps } = props;
+
+  return (
+    <FormControlLabel
+      id={`${props.fieldName}-label`}
+      control={
+        <CheckBox
+          id={`${props.fieldName}-cb`}
+          onChange={event => onChange(event.target.value)}
```
dawehner: Note: for setting the right checked value, you need to access ```event.target.checked```:

```
event.target.value
""
event.target.checked
true
```

rsanzante: Interesting, ```value``` returns "true" or "", while ```checked``` returns ```true``` or ```false```.

Changed!

dawehner:
@rsanzante Great work!

dawehner:
Do you want to come to our initiative meeting later in #javascript ?

Issue #157, opened 2018-07-04
Date selection field widget


dawehner:
Just like ```core/lib/Drupal/Core/Datetime/Plugin/Field/FieldWidget/TimestampDatetimeWidget.php:25``` we need a widget for timestamps

Right now the UI looks like this in Drupal 
![screen shot 2018-07-04 at 14 57 01](https://user-images.githubusercontent.com/29678/42281123-89a7f714-7f9a-11e8-91f9-2dc3cfadba3f.jpg)

There is a material UI component, see https://material-ui.com/demos/pickers/ we could use

VD39:
I will work on this issue.

dawehner:
This is done now!

PR #167, opened 2018-07-05
Added TimestampDatetime component

VD39:


src/components/05_pages/NodeForm/index.js
```diff
@@ -63,7 +63,7 @@ class NodeForm extends React.Component {
               // @todo How do we handle cardinality together with jsonapi
               // making a distinction between single value fields and multi value fields.
               return React.createElement(this.props.widgets[widget], {
-                key: { fieldName },
+                key: fieldName,
```
VD39: Updated the key as I kept getting the following error:

Warning: Encountered two children with the same key, `[object Object]`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.

VD39:
Reference: #157

dawehner:
When trying this element out I saw one bug:

timestamp

src/components/05_pages/NodeForm/index.js
dawehner: +1

src/components/02_atoms/Widgets/TimestampDatetime.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import TextField from '@material-ui/core/TextField';
+import WidgetPropTypes from '../../05_pages/NodeForm/WidgetPropTypes';
+
+const TimestampDatetime = ({ fieldName, label, value }) => (
```
dawehner: You need to use onChange, see ```StringTextfield```

dawehner:
It works great now

PR #169, opened 2018-07-06
Add a vertical layout for the forms

dawehner:
screen shot 2018-07-06 at 09 19 26


src/components/05_pages/NodeForm/index.js
```diff
@@ -1,5 +1,6 @@
-import React, { Fragment } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
+import { withStyles } from '@material-ui/core';
```
justafish: Let's use emotion here

PR #171, opened 2018-07-06
Enable Node Save

dawehner:
This PR is blocked on having a author field and a image upload field.

When committing we should add @e0ipso as contributor, given he helped with fixing the POST


src/components/05_pages/NodeForm/index.js
```diff
@@ -15,8 +20,17 @@ schemaType = PropTypes.shape({
   properties: PropTypes.objectOf(lazyFunction(lazySchemaType)),
 }).isRequired;
 
+const styles = () => ({
+  container: {
+    display: 'flex',
```
martinfrances107: There is a little bug hiding here 

it is similar to #202 - single quotes surrounding text properties of CSS styles is invalid.
This applies to 'flex', 'wrap' and 'column'

I hope this helps.

dawehner:
Wait but isn't this one actual JS?

martinfrances107:
> Wait but isn't this one actual JS?

Sorry the other example I quoted was using string template back-ticks.. so my comment is just noise.

dawehner:
Yes we do have issues for those. I know Francesco was working on it

dawehner:
I created a new followup: https://github.com/jsdrupal/drupal-admin-ui/issues/265

Issue #154, opened 2018-07-04
options_select field widget


dawehner:
Similar to https://github.com/jsdrupal/drupal-admin-ui/pull/148/files#diff-4ed0700d6e4aa3489751ab950e7bbe5a we should have a select widget ```\Drupal\Core\Field\Plugin\Field\FieldWidget\OptionsSelectWidget```

I suppose we will need some level of settings of the allowed values, which probably should be exposed as part of the schema. This might be a tricky issue ...

VD39:
I will take this issue.

PR #175, opened 2018-07-06
Added OptionsSelect component

VD39:


VD39:
Reference: #154

src/components/05_pages/NodeForm/RecipeSchema.js
```diff
@@ -133,8 +133,8 @@ export default {
         field_difficulty: {
           type: 'string',
           title: 'Difficulty',
-          enum: ['easy', 'medium', 'hard'],
-          default: 'medium',
+          options: ['easy', 'medium', 'hard'],
+          defaultValue: 'medium',
```
VD39: Change values, as ones used are reserved, so using them is not possible when deconstructing objects.

dawehner: Mh, I fear we will not be able to change this output. This is a copy of what will come back from openapi.

VD39: Change made and I also updated the component.

src/components/02_atoms/Widgets/OptionsSelect.js
```diff
@@ -0,0 +1,61 @@
...
+  `,
+};
+
+const OptionsSelect = ({
+  label,
+  value,
+  onChange,
+  helpText,
+  fieldName,
+  enum: options,
```
dawehner: Mh, this doesn't work yet because I guess we don't pass the schema along, yet

dawehner: Gotcha. I think we should pass along the schema of the field to each widget. This way you could extract the information you need. Does that make sense?

dawehner: Do you want to do this as part of this PR?

VD39: Sure, I can do that.

src/components/02_atoms/Widgets/OptionsSelect.js
```diff
@@ -0,0 +1,64 @@
...
+  schema: PropTypes.shape({
+    default: PropTypes.string,
+    enum: PropTypes.arrayOf(PropTypes.string),
+  }),
+};
+
+OptionsSelect.defaultProps = {
+  value: '',
+  helpText: '',
+  schema: {
```
dawehner: I do have one question: Why would you provide a default schema, which isn't really helpful. I'd rather suggest to make it required, so you cannot use it wrong.

VD39: Ah, fair point. Change has been made.

Issue #178, opened 2018-07-07
Make content page pass the lighthouse accessibility test


martinfrances107:
This is just a minor follow up to  #137 

When I run localhost:3000 through Chrome's Lighthouse.

I get two accessibility errors  related to  buttons ... inside TablePagination.
That they  do not have a aria-label or equivalent

From lighthouse you can click on a link to get more information 

https://dequeuniversity.com/rules/axe/2.2/button-name?application=lighthouse

TablePagination has the following properties 

backIconButtonProps
nextIconButtonProps

https://material-ui.com/api/table-pagination/

For a page reader, the pagination at the bottom on content.js can be tailored to what is on display, nothing fancy maybe 

Previous content / Next content

martinfrances107:
This is a todo list.

#201 and #196 -- fixed most of the stuff.. 😄 

Regarding the content page ... Lighthouse's A11y score has moved from the low forties  to 83/100.

To get 100/100 we need to fix two issues....

1) The input element  inside the 'TITLE" textField needs a aria-label.
2) The actions element which is greyed out until something is selected....but it  is too greyed out in that it now fails the contrast ratio test.

I am working on a PR now.

martinfrances107:
In the last PR #263   I fixed only -- half the issues on the todo list.

I opened up  #264 so that we can get input from a graph designer on the best way of fixing todo issue 2.

martinfrances107:
There is a side issue for the outstanding issue.

Thank you for looking at this.

PR #179, opened 2018-07-07
Accessibility review for content page ( #178 )

martinfrances107:
Accessibility review for content page ( Errors seen in Chrome's lighthouse )

A) Link between Select#action and label was broken [ typo in htmlFor ]

B) Made table column 2 the label for the column select element.

C) Added aria-label to the buttons in the TablePagination element

See issue #178


martinfrances107:
To criticise my own work ...  I can see a problem .. this is not yet ready to merge I will fix tomorrow..

Looking at the newly added formLabel ... it changes the color, font-size and font-weight of the title text my bad sorry

martinfrances107:
Looking on the internet there is an  issues reporting that the lint error is itself erroneous.

Here is the general form 

<label htmlFor="firstName">First Name</label>
<input id="firstName" type="text" />

and here is the report, reporting the **bad** lint failure.

https://github.com/evcohen/eslint-plugin-jsx-a11y/issues/314

This project is small and rapidly evolving .. instead of getting tied up in making every thing perfect from day one ... I suggest forcing  the .eslintrc file to ignore the test,  with a comment as to why, and move on.

.eslintrc.json
```diff
@@ -15,6 +15,7 @@
     "jsx-a11y/anchor-is-valid": [ "error", {
       "aspects": [ "noHref" ]
     }],
+    "jsx-a11y/label-has-for": "Off",
```
mattgrill: I think this should be `["off"]`

martinfrances107: If Off is in danger of being parsed as and string and the in turn cast to  a boolean true,  do you mind if I change it to 0 [zero]?

Issue #109, opened 2018-05-22
Leverage tools like react-a11y in build process


mgifford:
I was reading through what WP is doing for automated testing here:
  https://humanmade.com/2018/05/17/automated-accessibility-testing-during-development/

Came across these react tool for accessibility:
  https://github.com/reactjs/react-a11y
  https://github.com/evcohen/eslint-plugin-jsx-a11y

Would be good to have better tools to test the builds. 

Again, automated tests only go so far, but they are better than no tests. 

Mike

justafish:
We're already using react-a11y in our automated tests. pa11y looks useful :+1: We could also write some of these tests in Nightwatch.

mgifford:
I think Pa11y will be more useful when it is easier to run axe-core from it https://github.com/pa11y/pa11y/issues/399#issuecomment-390623275

dylanb:
You can also run axe-core directly in Enzyme and Jest tests, there is also the https://github.com/dequelabs/react-axe module

martinfrances107:
So here is a developers users experience story...

While running the content page through Chrome's lighthouse test suite .. ... I picked out manually a few accessibility bugs ( #178 ) **periodic manual inspection is bad** 

When I use react-axe ... it automatically and are highlights my bugs in Chrome browsers console logs ... 

So I have a pull request I am about to create.... its is only a few lines of code to review.

mgifford:
Also worth looking at https://github.com/evcohen/eslint-plugin-jsx-a11y & https://www.previousnext.com.au/blog/automate-your-drupal-accessibility-testing-axe-and-nightwatchjs

martinfrances107:
Thanks I will work that into the next update here...

Sorry to talk a little PR housekeeping 

When/If  we starting lint checking a11y --- initially  this issue will fail hard ---

So this is just a note to say we could conceptually 'postpone' and  possibly merge this once #179 and #196 are complete.

martinfrances107:
So I have a judgement call we need to talk about here....

I don't particularly feel strongly about one way  or the other

At hand we  have a switch to set for the a11y lint ruleset "recommended" or "strict"

It is a new project, undergoing rapid development  so 'strict' might be too strong when so much code is being written and then rapidly rewritten.

So the choice I have gone for  is lots of warnings/errors dumped to the console  ( react-axe ) and the more lackadaisical 'recommended' ruleset for linting.

dawehner:
[...]

PR #192, opened 2018-07-10
Improve reporting of  a11y issues [lint checks and react-axe ] ( #109 )

martinfrances107:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/109

## Screenshot / UI changes

## Testing instructions

Using Chrome's devtools accessibility errors are highlights in the console window

A DX improvement to produce early failure, in regard to a11y.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,7 +1,8 @@
 {
   "extends": [
     "airbnb",
-    "plugin:prettier/recommended"
+    "plugin:prettier/recommended",
+    "plugin:jsx-a11y/recommended"
   ],
   "parser": "babel-eslint",
   "env": {
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "eslint-config-airbnb": "^16.1.0",
     "eslint-config-prettier": "^2.9.0",
     "eslint-plugin-import": "^2.7.0",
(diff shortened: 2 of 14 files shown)


martinfrances107:
Here is a screenshot showing axe's helpful warning 

react-axe

dawehner:
Is there a way to automate some of those failures, so nightwatch catch them?

martinfrances107:
> Is there a way to automate some of those failures,


@mgifford has provided the answer ( see the issue ) 

inshort -- if set set the ruleset to "strict" we can trigger errors while linting.

martinfrances107:
I need to explain the latest push to the PR
 
🎉 It is all change Monday, 
 The random failure has been resolved, 
  The dbLog page is no longer 404'ing.

So here I can fix the onChange for onBlur in the places highlighted by the test errors.

Thank you all....

lauriii:
We currently have some failures but after these gets resolved, it would be great if our CI could somehow test these as well.

martinfrances107:
> it would be great if our CI could somehow test these as well.

@lauriii 

Can I pin this issue to  ( #109 ) 

Referring to one of my comments ...  I am advocating  that all the messy failures are dumped into console pane as the  developer works - Once our technical debt is fixed they maybe we should go for "strict" over "recommended"   as we set the ruleset

Issue #202, opened 2018-07-16
Malformed CSS is corrupting the styling of the menubar


martinfrances107:
[Blush ] Oh menubar your scrollbars are showing.

If you look at the image to the left just about the watercress soup table row -- there is a visual blemish 

it is a unwanted horizontal scroll widget 
 

![menubadscrollbar](https://user-images.githubusercontent.com/1259969/42749441-9cee1dc2-88db-11e8-93f1-a0daae6af06a.png)

the css definition is corrupted by the single quotes directly around **hidden**

I am creating a PR to remove this .. I hope it is seen as quick fix.

```
  drawerPaperClose: css
    overflow-x: 'hidden';
    transition: none;
    width: 72px;
    transition: width 195ms cubic-bezier(0.4, 0, 0.6, 1) 0ms;
  
```

martinfrances107:
I have corrected a hidden bug in the test strategy.

Previously the errant scrollbar could be manipulated to show the button with the menubar drawer closed. So technically a waitForElementVisible() passed even with the button 'off' the screen.

Now with the scrollbar gone, this waitForElementVisible() issue has been tightened up .. in turn the tests have to be altered so that they open the menubar drawer before inspection.

martinfrances107:
merged 🎉 Thank you all for the reviews.

PR #203, opened 2018-07-16
Remove horizontal scrollbar widget from the menu bar.(#202)

martinfrances107:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/202

see main issue for the screen shot

Look at the main content page - see the scrollbars can be seen as visual blemish.
[to the left and just above the watercress table row. ]

Chrome's devtools were reporting the  'hidden' property was invalid --  hidden without the quotes is a valid property!

### DIFF
--- a/src/components/06_wrappers/Default/Default.js
+++ b/src/components/06_wrappers/Default/Default.js
@@ -137,7 +137,7 @@ styles = {
     transition: width 225ms cubic-bezier(0.4, 0, 0.6, 1) 0ms;
   `,
   drawerPaperClose: css`
-    overflow-x: 'hidden';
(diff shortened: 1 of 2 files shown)


martinfrances107:
So I have a question

> Some checks were not successful

looking at the night watch test ... it says it is failing 

but when you click the link it directs me to a passing tests.

so it looks like the test message it out of sync with reality.... is there any way of forcing a retest?

martinfrances107:
Thanks to all who looked at this yesterday, 

The solution .. now that the scrollbars have been removed ... the menubar is in a sense 'correctly' hidden. -  so the drawer must be opened before inspection.

Issue #207, opened 2018-07-18
Tests: Write a jest test for the AddContent component


slipo:

dawehner:
🎆

PR #208, opened 2018-07-18
Jest tests for AddContent component

slipo:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/207

## Screenshot / UI changes

screen shot 2018-07-18 at 10 06 47 am

## Testing instructions

`yarn run test:unit AddContent`

This component requires store. I chose to export the unconnected component to get around that requirement because I think it's cleaner than mocking store.


slipo:
I disabled the no-named-as-default eslint rule.

Some more info:
https://stackoverflow.com/questions/44437203/how-do-i-resolve-eslint-import-no-named-as-default

I still think this is the cleanest way to test connected components. Open to suggestions.

package-lock.json
```diff
@@ -0,0 +1,17944 @@
+{
```
lauriii: We are using yarn on this project. Would it be possible to delete this file and update the yarn lock file instead?

lauriii:
About the `import/no-named-as-default`; one approach I've seen taken is that there's a separate file that does connecting to higher order components and the redux state. Another option would be to rename the component to something else so that it wouldn't be the same as the default export.

Issue #173, opened 2018-07-06
Tests: Write a nightwatch test for the /node/add page


dawehner:
It would be nice to have some tests for the /node/add page, just ensuring it lists the expected content types.

PR #211, opened 2018-07-18
Nigthwatch test for AddContent page (node/add)

slipo:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/173

## Screenshot / UI changes

screen shot 2018-07-18 at 11 47 36 am

## Testing instructions

`yarn test:functional src/tests/Nightwatch/Tests/AddContent.js`

### DIFF
--- a/src/tests/Nightwatch/Tests/AddContent.js
+++ b/src/tests/Nightwatch/Tests/AddContent.js
@@ -4,25 +4,26 @@ module.exports = {
     browser
       .logUserIn()
       .relativeURL('/node/add')
-      .waitForElementVisible('[data-nightwatch="node-type-list"] a', 10000)
-      .getText(
+      .waitForElementVisible('[data-nightwatch="node-type-list"] a');
+
+    browser.expect
+      .element(
         '[data-nightwatch="node-type-list"] a:nth-child(1) span:first-child',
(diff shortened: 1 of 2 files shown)


src/tests/Nightwatch/Tests/AddContent.js
```diff
@@ -0,0 +1,28 @@
+module.exports = {
+  '@tags': ['AddContent'],
+  addContentRenders(browser) {
+    browser
+      .logUserIn()
+      .relativeURL('/node/add')
+      .waitForElementVisible('[data-nightwatch="node-type-list"] a', 10000)
```
dawehner: Instead for specifying 10000 for every single of those links I'm wondering whether we should change the default instead, see https://github.com/jsdrupal/drupal-admin-ui/pull/214

src/tests/Nightwatch/Tests/AddContent.js
```diff
@@ -0,0 +1,28 @@
...
+      .getText(
```
dawehner: I'm wondering why you picked this way instead of something like: ```  browser.expect.element('[data-nightwatch="node-type-list"] a:nth-child(1) span:first-child').text.to.equal('Article');
``` ❓

Issue #184, opened 2018-07-09
Create standardized way of rendering parts of the page


lauriii:
It would be useful to have standardized way of providing data from page level components so that it would be possible to render important parts of the page in a standardized way. Good example of this is the page title which we probably want always to exist and we want it to be always rendered the same way. There is probably other use cases as well, for example local actions and tasks.

One of the methods would be to use interfaces, which unfortunately doesn't exist in JavaScript. Using TypeScript would solve this #64.

Other option would be to use https://github.com/gaearon/react-side-effect library which allows components to trigger global side effects.

justafish:
Could you clarify a bit further what you mean here? I'm not understanding what the requirements are that aren't solved by having e.g. a PageTitle component.

lauriii:
Mainly this is about the fact that we want to always have the page title rendered in the same place on the page. If rendering title is done manually by components rendering pages, that might lead into inconsistency between pages. This is a pre-existing UI pattern in Drupal which we might want to bring into this application.

dawehner:
I guess a requirement we have is not only that we want the page title to always appear in the same place, it is more about there is always one.

Is the admin UX design team 100% sure that this is the pattern which will exist in the future?

On the implementation side of things I am wondering whether we could use the fact that react components can return not just elements, but also array of elements.

ckrina:
In the title situation specifically, there will be always a title but I'm not sure it will always behave the same way. In example, let's say you are creating a new node: now the title is something like "Create Basic page" but in the future it could be changed to something that represents the real title for new content you are creating.
And local actions could be extended by other modules maybe?
Maybe the link to the user profile could be a recurrent thing that doesn't change.

lauriii:
> Maybe the link to the user profile could be a recurrent thing that doesn't change.

This seems like something that doesn't change per page.

> In the title situation specifically, there will be always a title but I'm not sure it will always behave the same way. In example, let's say you are creating a new node: now the title is something like "Create Basic page" but in the future it could be changed to something that represents the real title for new content you are creating.

The question is that do we want to force every page to have a title as a UX pattern or not.

ckrina:
Yes, there should be a title to let people have context. But it might not look the same, be placed on the  same region or behave the same.

PR #215, opened 2018-07-19
Add PageTitle component

lauriii:
## Screenshot / UI changes

This create `PageTitle` component which has a side effect to change the document title. This takes into accounts requirements set on https://github.com/jsdrupal/drupal-admin-ui/issues/184.

A downside on the current approach compared to what I proposed as an alternative there is that we cannot pre-fetch the title from the component, meaning that initially the page will render using wrong page title.

## Testing instructions

Content related pages should have titles, and when navigating between pages the document title should change.

### DIFF
--- a/src/components/02_atoms/PageTitle/PageTitle.js
+++ b/src/components/02_atoms/PageTitle/PageTitle.js
@@ -1,0 +1,36 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import withSideEffect from 'react-side-effect';
+import { css } from 'emotion';
+
+import Typography from '@material-ui/core/Typography';
+
+const styles = {
+  title: css`
+    margin-bottom: 15px;
+    margin-left: 3px;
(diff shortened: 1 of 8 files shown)


src/components/02_atoms/PageTitle/PageTitle.js
```diff
@@ -0,0 +1,40 @@
...
+  <Typography variant="headline" classes={{ root: styles.title }}>
+    {children}
+  </Typography>
+);
+
+PageTitle.propTypes = {
+  children: PropTypes.node.isRequired,
+};
+
+function reducePropsToState(propsList) {
```
dawehner: What is the reason to not use ```const = ```functions like everywhere else?

lauriii: Good point! 👍 This was leftover from a copy and paste from the example 🙂

PR #216, opened 2018-07-19
Update admin_ui_support dependencies

tedbow:
## Issue

We need these for testing

### DIFF
--- a/admin_ui_support/composer.json
+++ b/admin_ui_support/composer.json
@@ -10,7 +10,7 @@
         "phpmetrics/phpmetrics": "^2.3"
     },
     "require": {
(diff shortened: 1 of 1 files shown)


admin_ui_support/composer.json
```diff
@@ -10,6 +10,7 @@
         "phpmetrics/phpmetrics": "^2.3"
     },
     "require": {
-        "drupal/jsonapi": "^1.15"
+        "drupal/jsonapi": "^1.19",
```
mattgrill: Should this be `1.22` ?

tedbow: Yes, updated(though "^" would have got us this anyways)

Issue #155, opened 2018-07-04
Entity reference field widget


dawehner:
We need some entity reference field widgets for

* Referencing users
* Referencing taxonomy terms, creating tags

There is https://material-ui.com/demos/autocomplete/#autocomplete available, but it wraps other libraries. It would be nice to look at those libraries and see whether they allow to create new tags as well.

* Creating tags might be also tricky as the API might not support it yet directly.

An additional widget is available on https://github.com/jsdrupal/drupal-admin-ui/pull/148/files#diff-4ed0700d6e4aa3489751ab950e7bbe5a

francescopersico:
I 'll work on the issue.

PR #218, opened 2018-07-19
Entity Reference Widget

mattgrill:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/155

## Screenshot / UI changes

react_app

## Testing instructions

- visit `/node/add/recipe`

## Todo

- [x] cleanup keyboard actions, things a _bit_ buggy.
- [x] pass selected items back to the parent
- [ ] save the entity reference information along with the node.


src/components/02_atoms/Widgets/EntityReference.js
```diff
@@ -0,0 +1,186 @@
...
+
+    this.setState({
+      inputValue: '',
+      selectedItem,
+    });
+  };
+
+  handleInputChange = event => {
+    this.setState({ loading: true, inputValue: event.target.value }, () => {
+      api('taxonomy_term', {
```
larowlan: should this be a prop for greater reusability when you need to reference users too?

src/components/02_atoms/Widgets/EntityReference.js
```diff
@@ -0,0 +1,186 @@
...
+              condition: {
+                path: 'name',
+                operator: 'CONTAINS',
+                value: this.state.inputValue,
+              },
+            },
+          },
+        },
+        parameters: {
+          termType: this.props.inputProps.resourceIdentifer,
```
larowlan: same

src/components/02_atoms/Widgets/EntityReference.js
```diff
@@ -0,0 +1,186 @@
...
+        },
+      }).then(({ data: terms }) => {
```
larowlan: same we could call this items for portability

src/components/02_atoms/Widgets/EntityReference.js
dawehner: yeah this totally makes sense as next step :)

src/components/02_atoms/Widgets/EntityReference.js
dawehner: question: How do we know which field to filter in a generic widget?

src/components/02_atoms/Widgets/EntityReference.js
mattgrill: Yes, this could easily come from uiMetadata

src/components/02_atoms/Widgets/EntityReference.js
mattgrill: Could be part of the constraints un uiMetadata.

src/components/02_atoms/Widgets/EntityReference.js
mattgrill: We should probably actually break this entire fetching bit into its own file at some point.

src/components/05_pages/NodeForm/UiMetadata.js
```diff
@@ -44,6 +44,13 @@ const UiMetadata = {
       suffix: ' minutes',
     },
   },
+  field_tags: {
```
lauriii: We should at least open a follow-up to support auto creating entities. The recipes content type doesn't use that on the tags field but the articles content type does.

src/components/02_atoms/Widgets/EntityReference.js
```diff
@@ -0,0 +1,186 @@
...
+    this.setState({
+      inputValue: '',
+      selectedItem,
+    });
+  };
+
+  handleInputChange = event => {
+    this.setState({ loading: true, inputValue: event.target.value }, () => {
+      api('taxonomy_term', {
+        queryString: {
```
lauriii: The current architecture abstracts the querying part outside of the widget. Should we open a follow-up to figure out if we have to implement that on the client side as well?

dawehner:
I addressed points I had about this PR in some commits.

src/components/02_atoms/Widgets/EntityReferenceAutocompleteWidget.js
```diff
@@ -0,0 +1,215 @@
...
+import Paper from '@material-ui/core/Paper';
+import TextField from '@material-ui/core/TextField';
+import Chip from '@material-ui/core/Chip';
+import MenuItem from '@material-ui/core/MenuItem';
+
+import WidgetPropTypes from '../../05_pages/NodeForm/WidgetPropTypes';
+
+import api from './../../../utils/api/api';
+
+class EntityReferenceAutocompleteWidget extends React.Component {
```
lauriii: We probably shouldn't include `Widget` in the name since most of the widgets doesn't have it.

dawehner:
@lauriii Regarding the error, see https://github.com/jsdrupal/drupal-admin-ui/issues/221#issuecomment-406948175

src/components/02_atoms/Widgets/EntityReferenceAutocompleteWidget.js
dawehner: Sure, let's do that!

dawehner: Do you want to put up a PR for the file upload widget?

mattgrill:
@lauriii @dawehner I just fixed the issue with the bad object-key reference for the schema.

src/components/02_atoms/Widgets/EntityReferenceAutocomplete.js
```diff
@@ -0,0 +1,222 @@
...
+    ...WidgetPropTypes,
+    inputProps: PropTypes.shape({
+      bundle: PropTypes.string,
+      type: PropTypes.string,
+    }).isRequired,
+  };
+
+  state = {
+    inputValue: '',
+    selectedItems: {},
```
mattgrill: This can probably be a `Map`, but I am sort of lazy.

mattgrill:
@dawehner Based on our conversation in Slack, I am ➕1️⃣ on merging this and handling the remaining tasks either in the node-save issue or followups.

PR #219, opened 2018-07-20
Document jest vs. nightwatch

dawehner:
## Issue

Let's document when to use jest vs. nightwatch


README.md
```diff
@@ -118,3 +118,11 @@ If someone has made a pull request and you would like to add code to their branc
   .
   git push <your-fork> <branch-name>
   ```
+  
+  ## Development best practises
+  
+  ### Testing
+  
+  We do have both functional level testing ([nightwatch](http://nightwatchjs.org/)), but also more component/unit testing ((jest)[https://jestjs.io/]).
```
justafish: Improving the English:

We have functional testing with [nightwatch](http://nightwatchjs.org/), and component/unit testing with [jest](https://jestjs.io/).

README.md
```diff
@@ -118,3 +118,11 @@ If someone has made a pull request and you would like to add code to their branc
...
+  
+  If you struggle to decide where to add test coverage: Nightwatch should test the common path and jest is for detailed test coverage.
```
justafish: When deciding which system to use to add test coverage, the general rule is Nightwatch should be used to test the common path, and Jest can be used for more detailed test coverage. Nightwatch tests will run slower as they simulate clicking around in a real browser.

dawehner: 👍

README.md
dawehner: 👍

justafish:
- There's still some odd brackets in here and the jest link is broken
- Capitalise Nightwatch and Jest
- There's some spaces in front of the h2s an h3s which are making the indentation weird
- Needs a menu entry at the top
- Maybe let's move this above the forking instructions?

dawehner:
🤦‍♂️

PR #222, opened 2018-07-20
Fetch Schema Automatically

mattgrill:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/221

## Screenshot / UI changes

No UI changes, just fetching the schema automatically.

## Testing instructions

Ensure that you have the latest support module enabled, and openapi/schemata. Ensure that `/node/add/recipe` works as expected.

PR #227, opened 2018-07-23
Fix schema fetching

dawehner:
Resolves a small regression introduced by https://github.com/jsdrupal/drupal-admin-ui/pull/222


src/components/05_pages/NodeForm/index.js
```diff
@@ -70,8 +70,8 @@ class NodeForm extends React.Component {
   };
 
   getSchemaInfo = (schema, fieldName) =>
-    this.props.schema.properties.attributes.properties[fieldName] ||
-    this.props.schema.properties.relationships.properties[fieldName];
+    this.props.schema.properties.data.properties.attributes.properties[fieldName] ||
```
francescopersico: You pass schema but use this.props....

PR #229, opened 2018-07-23
Improve steps used to setup Drupal for integration tests

lauriii:
This PR changes the Drupal installation commands for the integration test. This tries to tackle various problems at once:
1. The commands run cloned git repositories not used for testing which made following the installation steps slightly more complex than it could be.
2. Drupal was installed using the latest dev code, not the actual code being tested. This could have caused failures if we changed the installation requirements (not sure if this is the case since the tests install Drupal separately always).
3. As a side benefit this should also make the tests run a bit faster

We do need a follow-up to add usages of the commands to the demo repository to make sure all commands are being at least somewhat tested.


mattgrill:
👍

Issue #223, opened 2018-07-21
Fix error on booleanCheckbox


dawehner:
Right now there is the following error message:

```
Warning: React does not recognize the `entityTypeId` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `entitytypeid` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
    in span (created by ButtonBase)
    in ButtonBase (created by WithStyles(ButtonBase))
    in WithStyles(ButtonBase) (created by IconButton)
    in IconButton (created by WithStyles(IconButton))
    in WithStyles(IconButton) (created by SwitchBase)
    in SwitchBase (created by WithStyles(SwitchBase))
    in WithStyles(SwitchBase) (created by Checkbox)
    in Checkbox (created by WithStyles(Checkbox))
    in WithStyles(Checkbox) (at BooleanCheckbox.js:14)
    in label (created by FormControlLabel)
    in FormControlLabel (created by WithStyles(FormControlLabel))
    in WithStyles(FormControlLabel) (at BooleanCheckbox.js:11)
    in BooleanCheckbox (created by NodeForm)
    in form (at index.js:86)
    in div (created by Paper)
    in Paper (created by WithStyles(Paper))
    in WithStyles(Paper) (at index.js:85)
    in NodeForm (created by Connect(NodeForm))
    in Connect(NodeForm) (at routes.js:19)
    in nodeAddRecipe (created by Route)
    in Route (created by withRouter(nodeAddRecipe))
    in withRouter(nodeAddRecipe) (created by Route)
    in Route (at App.js:54)
    in Switch (at App.js:51)
    in ErrorBoundary (at Default.js:106)
    in main (at Default.js:105)
    in div (at Default.js:55)
    in Default (created by Connect(Default))
    in Connect(Default) (at App.js:50)
    in Router (created by ConnectedRouter)
    in ConnectedRouter (at App.js:49)
    in Provider (at App.js:48)
    in ErrorBoundary (at App.js:47)
    in JssProvider (at App.js:46)
    in App (at index.js:12)
```

dawehner:
Thank you for the research of what is going. That's not obvious at all. Your fix seems totally sensible

I'm wondering whether @justafish @mattgrill or @lauriii have an opinion here.

lauriii:
IMO we shouldn't ever pass this prop to the child since it is totally unrelated.

mattgrill:
[...]

PR #230, opened 2018-07-23
Don't render additional widget props as HTML field attributes

lhuria94:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/223

## Screenshot / UI changes

No UI changes.

## Testing instructions
The unknown-prop warning will fire if we attempt to render a DOM element with a prop that is not recognized by React as a legal DOM attribute/property.

This happened on `/node/add/recipe` page where we are sending `entityTypeId` to all of its widget components For example: `booleanCheckbox, FileUploadWidget` etc.

- Navigate to `/node/add/recipe`, what we see now is this `Warning: React does not recognize the `entityTypeId` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `entitytypeid` instead. If you accidentally passed it from a parent component, remove it from the DOM element.`.
- After pulling this branch, the error/Warning should go away, as we are assigning the props to a new object and deleting the keys which we do not need.


src/components/02_atoms/Widgets/BooleanCheckbox.js
```diff
@@ -5,7 +5,9 @@ import FormControlLabel from '@material-ui/core/FormControlLabel';
 import WidgetPropTypes from '../../05_pages/NodeForm/WidgetPropTypes';
 
 const BooleanCheckbox = props => {
-  const { onChange, label, id, value, ...checkboxProps } = props;
+  const { onChange, label, value } = props;
+  const checkboxProps = Object.assign({}, props);
```
mattgrill: What about doing something like, `const { entityTypeId, ...checkboxProps } = props;`?

src/components/02_atoms/Widgets/BooleanCheckbox.js
```diff
@@ -5,7 +5,9 @@ import FormControlLabel from '@material-ui/core/FormControlLabel';
...
+  delete checkboxProps.entityTypeId;
```
lauriii: I still think this should be done on a higher level - we never should have passed this prop in the first place. I think this will get out of control if we don't control the props set for widgets.

src/components/02_atoms/Widgets/BooleanCheckbox.js
lhuria94: Yes, it's much better. :) Thanks

src/components/02_atoms/Widgets/BooleanCheckbox.js
lhuria94: Makes sense though, that's what I mentioned in the issue as well. If we are not going to be able to use, then why pass it.

lauriii: Exactly 👍

lhuria94: Need to double check though if there is any impact, because if we might need to remove this prop from here: https://github.com/jsdrupal/drupal-admin-ui/blob/master/src/components/05_pages/NodeForm/index.js#L101

And this will have it removed from all of the widgets where FileUploadWidget seem to be using this as parameter: https://github.com/jsdrupal/drupal-admin-ui/blob/master/src/components/02_atoms/FileUpload/FileUpload.js#L56

lauriii: We could create object with key `metadata` or something similar that could contain this type of information. 🙂

dawehner:
We neeed to know it for the file upload widget

lauriii:
Actually looking at this more closely it seems like the best solution would be to never spread the object to the form element. This way, if future props gets added there shouldn't be any problems.

Issue #235, opened 2018-07-24
 Give OpModalButton a story book page


martinfrances107:
Acting on a suggestion from @dawehner 

I am going to split off a issue from https://github.com/jsdrupal/drupal-admin-ui/issues/186
and just provide a storybook page for the OpModalButton.

PR Is comming

martinfrances107:
Thank you ❤️

Issue #186, opened 2018-07-09
Give all atoms a story book page


martinfrances107:
Story book is new (  #183 ) 

Regarding the opModalButton component I want a story book entry, if only because it documents/details to new developers how icons are expected to be added as children

I have it already to go ... I just need a issue to associate a PR with.

martinfrances107:
I have widened this issue to cover more of what react calls  "atoms" 

I have updated the storybook page for OpsModalButton so that is now shows the button in situ... inside a Table element.

Now all elements in the widget sub directory have a storybook page.

dawehner:
@martinfrances107 I'd recommend to not try to expand the PR but rather do one thing at a time.

martinfrances107:
I need to find more time to work on this.. Yes OpsModalButton could be split off into a isolated PR. 

Regarding the widgets - to meet the next developers expectations, I think the next step is to have a single page for each widget and then have a "contact sheet" where all widgets are displayed all at once. 

But I have been opening issues where I see problems .. and I am now juggling too many minor issues, and so the next next step is to close the other loose ends.

martinfrances107:
❤️

martinfrances107:
sorry closed the wrong issue.

martinfrances107:
Thank you ❤️

PR #236, opened 2018-07-24
#186 Add storybook for OpsModalButton wrapped in a table.

martinfrances107:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/235

## Testing instructions

run:-

yarn storybook

See new page for the OpsModalButton

The button is on the right.

### DIFF
--- a/src/components/02_atoms/OpsModalButton/OpsModalButton.stories.js
+++ b/src/components/02_atoms/OpsModalButton/OpsModalButton.stories.js
@@ -11,7 +11,19 @@ import TableRow from '@material-ui/core/TableRow';
 
 import OpsModalButton from './OpsModalButton';
 
-storiesOf('Table Operations/OpsModalButton', module)
+storiesOf('OpsModalButton', module)
+  .add('Default', () => (
+    <OpsModalButton
+      aria-label="edit"
+      title="Are you sure that you want to edit this element?"
+      text="This action cannot be undone."
+      cancelText="Cancel"
+      confirmText="Edit"
+      enterAction={() => {}}
+    >
+      <EditIcon />
+    </OpsModalButton>
+  ))
   .addDecorator(story => (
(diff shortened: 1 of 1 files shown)


src/components/02_atoms/OpsModalButton/OpsModalButton.stories.js
```diff
@@ -0,0 +1,43 @@
...
+import EditIcon from '@material-ui/icons/Edit';
+import Table from '@material-ui/core/Table';
+import TableBody from '@material-ui/core/TableBody';
+import TableCell from '@material-ui/core/TableCell';
+import TableHead from '@material-ui/core/TableHead';
+import TableRow from '@material-ui/core/TableRow';
+
+import OpsModalButton from './OpsModalButton';
+
+storiesOf('Table Operations/OpsModalButton', module).addWithJSX(
```
lauriii: Couldn't this component be used outside of tables? 🤔

martinfrances107: > Couldn't this component be used outside of tables?

Maybe, 

The Ops in OpsModalButton - is short for 'object in the operations column of a table'.

I am still thinking of what are the next steps.... I could put the table as a decorator so when a user reads the bottom section of the OpsModalButton he/she is not immediately distracted.

Or maybe there would be two pages one showing the button in isolation and one with the button in the table? I will look at this some more and fix this up tonight.

martinfrances107:
See below I had made the storybook easier to grasp, using the addDecorator() method the JSX tab just contains the OpsModalButton element.


hiddentable

martinfrances107:
After a chat with lauriii....

I have split the representation of the button into two storybook pages.

1) The button in isolation.
2) The button contained in the operations column of a table.

Issue #199, opened 2018-07-13
Discuss the future development of Dblog


martinfrances107:
From #198 - dawehner 

> Whether we should maybe park the dblog implementation for now, given we want to focus on other parts?

So this is just an issue to organise the future development of the dblog page.

dawehner:
From my point of view we will not put our efforts onto this page for now. Given that I think we should drop it, git allows us to always get back the work, if we need it.

lauriii:
+1 on removing the dblog page. We are not utilizing these pages anywhere at the moment and also they won't be part of our demos. Also, we won't be getting designs for these pages in a while. Once we are at a point that we want to implement these, we can always look back to what we did with git.

dawehner:
@martinfrances107 Do you want to create a PR to drop it?

PR #238, opened 2018-07-25
Remove dblog and permissions pages since they are not supported at the moment

lauriii:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/199


mattgrill:
Could we just remove the route overrides from the support module and the route matches on JS side? Seems a bit much to delete everything.

lauriii:
I agree that it is a hard choice to delete these pages from the app now. I think it's better to do delete them completely from the app because we haven't been actively testing those pages and they also don't implement material UI. Maintaining those pages inside the app would add some extra cost with a little to none benefit at this point. Once we are getting more stable, we should just add them back and update to the latest APIs.

dawehner:
Keeping unused code alround is IMHO always worse than removing it. People wonder, try to fix things etc. Getting the page back would be overall less work, I think, when we completely remove it now.

dawehner:
@justafish Do you have an opinion about this?

justafish:
@lauriii the issue you linked is just for removing the dblog page, which I think is fine. Permissions works and is definitely a UX improvement even if it doesn't use Material UI yet, so let's leave that in.

mattgrill:
@justafish - I think we did just fix the dblog page.

> Keeping unused code alround is IMHO always worse than removing it.

I agree. I guess I don't have a real problem with removing some stuff for the sake of the demo.

admin_ui_support/admin_ui_support.routing.yml
```diff
@@ -1,48 +1,4 @@
 # Routes controlled by the React App.
-admin_ui_support.user.permissions:
```
dawehner: This certainly removes the permissions page as well

dawehner:
> I agree. I guess I don't have a real problem with removing some stuff for the sake of the demo.

How about writing a test to ensure it always works?

admin_ui_support/admin_ui_support.routing.yml
mattgrill: Maybe we can just remove the dblog component here and have an issue to update the permissions/roles pages to the material ui design?

dawehner: Sounds sensible for me!

dawehner:
@lauriii Are you planning to come back to that or should we tag this issue as "good first issue"?

justafish:
Looks like no one's working on this, so let's close for now and address when we do our clean up/housekeeping during around Drupal Europe

Issue #237, opened 2018-07-24
Enable the openapi module


VD39:
It seems the openapi module doesn't get enabled, causing an 404 API error when trying to add a recipe. 

Needs to be added here: https://github.com/jsdrupal/drupal-admin-ui/blob/master/admin_ui_support/admin_ui_support.info.yml

dawehner:
> Needs to be added here: https://github.com/jsdrupal/drupal-admin-ui/blob/master/admin_ui_support/admin_ui_support.info.yml

+1 Do you want to create a PR for that?

VD39:
Pull request opened. I am new this area, so I am not sure if it is correct.

lauriii:
Good catch! This definitely is a problem. Thank you for opening the PR!

VD39:
Closing as issue is fixed.

PR #239, opened 2018-07-25
Added openapi to list of dependencies

VD39:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/237


lauriii:
Do you think we would be able to add a very simple test for this bug?

dawehner:
Once we would have https://github.com/jsdrupal/drupal-admin-ui/issues/173 in place this test wouldn't be needed anymore. But yeah more tests often don't cause harm.

PR #246, opened 2018-07-26
Fix document.title formatting

mattgrill:
## Screenshot / UI changes

create_recipe

### DIFF
--- a/src/components/02_atoms/PageTitle/PageTitle.js
+++ b/src/components/02_atoms/PageTitle/PageTitle.js
@@ -31,6 +31,12 @@ const reducePropsToState = propsList => {
   return false;
 };
 
+/**
+ * Set the title of the page based on the children of PageTitle.
+ *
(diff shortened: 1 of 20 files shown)


src/components/02_atoms/PageTitle/PageTitle.js
```diff
@@ -32,7 +32,8 @@ const reducePropsToState = propsList => {
 };
 
 const handleStateChangeOnClient = title => {
-  document.title = title || '';
+  document.title =
```
dawehner: I'm wondering why do we allow an array as title in the first place? Shouldn't we disallow that?

mattgrill:
```

dawehner: Do you mind documenting this in the code?

PR #248, opened 2018-07-27
Hide react-axe behind an environment variable.

mattgrill:
## Issue

`react-axe` causes some outstanding performance issues, and a nasty call-stack error when sorting the content listing table.

react-axe

## Screenshot / UI changes

Disable `react-axe` unless you specify a environment variable, `REACT_APP_AXE=true yarn start`

## Testing instructions

Start without `react-axe` and be amazed how fast everything is.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,13 @@ cd drupal-admin-ui
 yarn start
 ```
 
+### React AXE
+
+Due to outstanding performance issues, `react-axe` is behind a flag. To enable the assessment provided by `react-axe` pass an environment variable when starting the application.
+
+```
+REACT_APP_AXE=true yarn start
+```
 
 ## Commands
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -70,6 +70,13 @@ cd drupal-admin-ui
 yarn start
 ```
 
+### React AXE
+
+Due to outstanding performance issues, `react-axe` is behind a flag. To enable the assessment provided by `react-axe` pass an environment variable when starting the application.
+
+```
+REACT_APP_AXE=true yarn start
```
dawehner: Interesting!

Should we maybe set this environment variable as part of nightwatch? I'm wondering whether we want to log the errors as part of testing or not.

mattgrill: Yes. We could probably enable this on the CI environment. However, `react-axe` appears to add a pretty significant performance penalty which might have been some of the timeout nightwatch issues.

Issue #247, opened 2018-07-27
FileUpload does not handle terminated uploads with grace.


martinfrances107:
There is a minor corner case outcome which is not being trapped and handled by our code.

There are several FileUpload issues

(https://github.com/jsdrupal/drupal-admin-ui/issues/193) is adding enhancements.
(https://github.com/jsdrupal/drupal-admin-ui/issues/190) shows that flow was used to catch the bug.
(https://github.com/jsdrupal/drupal-admin-ui/issues/186) is adding a storbook page for the widget.

The bug is in the FileUpload **ondrop** handler the const **files** can be null and yet "files" is always accessed as a iterator object. ( see 190 for details ).

I don't think this is high priority, I am juggling a few more important issues -- but I intend to come back to this.

dawehner:
Thank you for documenting processes like this. People don't do that often
enough, while it gives so much longterm benefit!
I'll review PRs as you go

VD39:
@martinfrances107 I am working on file upload at the moment, I am done with file upload component, which support multiple, and I am now working on the widget to get it to also support multiple uploads. So I can add the error handling or would you prefer to do it once I complete issue #193?

martinfrances107:
> So I can add the error handling or would you prefer to do it once I complete issue

@VD39 

I am struggling with time at the moment, and in addition I have a few too many issues open at the moment.

So I am easy ... with a community spirited attitude - as you post a PR over there  I will comment/review and help out where I can. Also if you choose to ignore this little bug .. I am good with that.

dawehner:
@martinfrances107 it would be super useful if you could try it out again, once #249 is in.

martinfrances107:
Sorry for the delayed response 

So the original problem :-

>  FileUpload ondrop handler the const files can be null and yet "files" is always accessed as a iterator object. ( see 190 for details ).

in the updated version event.dataTransfer.files is blindly passed to the function readFiles() where it hits this code.

```
  const slicedFiles: Object = Object.keys(files)
      .slice(0, 2)
      .reduce((obj, value) => {
        obj[value] = files[value];
        return obj;
      }, {});
```

when files is null javascript will throw this exception 

> Uncaught TypeError: Cannot convert undefined or null to object

So the fix appears simple -- add an "if" here and message the user that the transfer failed. 

```
  /**
   * Will get the selected file/s from the file explorer.
   * @param {Event} event
   */
  getFiles = (event) => {
    files = event.target.files
[...]

Issue #193, opened 2018-07-10
Enhancements to the file upload widget


dawehner:
* [ ] Allow to click into the drag zone to upload a file, not just drag'n drop
* [ ] Allow multiple files to be uploaded at the same time

VD39:
I will work on this.

VD39:
@dawehner should the images also show? At the moment it shows only the name for me.

dawehner:
That would be nice. Please ask ckristina whether there are any designs
available

dawehner:
Yeah, thank you @VD39 !

PR #249, opened 2018-07-30
Various updates to file upload widget

VD39:
## Issue 

https://github.com/jsdrupal/drupal-admin-ui/issues/193 and https://github.com/jsdrupal/drupal-admin-ui/issues/247

## Screenshot / UI changes

image
image

- Allow multiple and single file uploads.
  - If single the upload option will be set to display none so no more images can be uploaded.
  - Multiple will remain and allow for as many images to be uploaded.
- Added a button to upload area, this is so we have some accessibility for now. Will need to come back to this.
  - If single file upload and more than one file is dropped, it will try to upload the first image, if there is an error this will display instead.
- Error check before uploading.
  - Errors will display for every file that did not get uploaded.
  - Error text taken from Drupal.
- Add the option to delete a file.
- Image preview, used {{process.env.REACT_APP_DRUPAL_BASE_URL}} to be able to view the image.
- Settings for the inputProps were taken from: https://github.com/drupal/drupal/blob/e81312ba480f77becccb29bfb0bc5a7d311aeea5/core/profiles/demo_umami/config/install/field.field.node.recipe.field_image.yml#L19, not sure if this is correct but changing this should be simple.
- I added a multiple prop - this will need to change of course, I couldn't find where this option will come from.
- Not used to styled components, so not sure if it is 100% correct.
- Commented code to help better understand what is happening.
- Disabling the drag and drop while uploading, we don't queue the uploads so will have to disable to prevent the state from clearing and stopping the upload progress.

### What is missing:

- What is in the // TODO: 
- Other settings for error checking like max uploads, and max and min resolutions.
- Styles - Will communicate with ckristina for the design.
- Reorder of images.

## Testing instructions

Update the input props here: https://github.com/jsdrupal/drupal-admin-ui/pull/249/files#diff-71e62d02975664ca080c172b1b65cbf7R168

Can test the size, allowed files, and if multiple and single files.

### DIFF
--- a/src/components/02_atoms/FileUpload/FileUpload.js
+++ b/src/components/02_atoms/FileUpload/FileUpload.js
@@ -3,8 +3,10 @@ import PropTypes from 'prop-types';
 import styled from 'react-emotion';
 import Card from '@material-ui/core/Card';
 import Button from '@material-ui/core/Button';
+import Typography from '@material-ui/core/Typography';
 import CardContent from '@material-ui/core/CardContent';
 import CloudUploadIcon from '@material-ui/icons/CloudUpload';
+import LinearProgress from '@material-ui/core/LinearProgress';
 import api from './../../../utils/api/api';
 
 const Element = styled('div')`
@@ -13,6 +15,7 @@ const Element = styled('div')`
   padding: 80px;
   width: 100%;
   position: relative;
+  margin-bottom: 20px;
 `;
 
 const Text = styled('div')`
@@ -28,19 +31,13 @@ const Text = styled('div')`
   }
 `;
 
-const Error = styled('div')`
-  color: red;
+const marginBottom = {
+  marginBottom: '20px',
+};
 
-  > ul {
-    > li {
-      margin-bottom: 10px;
-    }
-
-    p {
-      margin: 0;
-    }
-  }
-`;
+const error = {
+  color: 'red',
+};
 
 class FileUpload extends Component {
   static propTypes = {
@@ -48,8 +45,9 @@ class FileUpload extends Component {
     bundle: PropTypes.string.isRequired,
     fieldName: PropTypes.string.isRequired,
     onFileUpload: PropTypes.func.isRequired,
+    multiple: PropTypes.bool.isRequired,
+    RemainingUploads: PropTypes.number.isRequired,
     inputProps: PropTypes.shape({
-      multiple: PropTypes.bool,
       file_extensions: PropTypes.string,
       max_filesize: PropTypes.number,
     }).isRequired,
@@ -63,6 +61,7 @@ class FileUpload extends Component {
     files: [],
     errors: [],
     filesLength: 0,
+    isDisabled: false,
   };
 
   /**
@@ -73,7 +72,7 @@ class FileUpload extends Component {
   onDragOver = event => {
     event.stopPropagation();
     event.preventDefault();
-    this.setBorder('red');
+    this.setElementStyles('red');
     event.dataTransfer.dropEffect = 'dragend';
   };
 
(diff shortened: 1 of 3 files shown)


src/components/02_atoms/FileUpload/FileUpload.js
```diff
@@ -1,62 +1,155 @@
...
+        const {
+          resetState,
+          props: { entityTypeId, bundle, fieldName, onFileUpload },
+        } = this;
         const { buffer } = new Uint8Array(result);
-
         const token = await api('csrf_token');
+        // Replace file name, some reason any space doesn't work
+        // TODO: Find a way to fix this without changing the name
+        const fileName = file.name.replace(/[,#!$^&*;{}=\-+`~()[\] ]/g, '_');
```
dawehner: Did you try using ```encodeURIComponent``` for the filename inside api.js? It looks like encoding the filename there could fix the problem you ran into

src/components/02_atoms/FileUpload/FileUpload.js
```diff
@@ -67,34 +160,187 @@ class FileUpload extends React.Component {
...
+
+  render = () => {
+    const {
+      onDrop,
+      onClick,
+      getFiles,
+      onDragOver,
+      onDragLeave,
+      state: { errors },
+      props: {
```
dawehner: Instead of multiple I would propose to look at the cardinality option on the schema. This way we would know whether you should be able to upload multiple files or not. This way we don't even need this parameter :) If cardinality > 1 then we should just always allow to upload multiple files.

src/components/02_atoms/FileUpload/FileUpload.js
```diff
@@ -67,34 +160,187 @@ class FileUpload extends React.Component {
...
+              style={{ display: 'none' }}
+              ref={element => {
+                this.input = element;
+              }}
+            />
+          </Element>
+
+          {errors.length > 0 && (
+            <Error>
+              <p>One or more files could not be uploaded.</p>
```
dawehner: Could we use some material UI components for displaying these error messages maybe?

src/components/02_atoms/FileUpload/FileUpload.js
VD39: This didn't work for me, maybe you could take a look?

src/components/02_atoms/FileUpload/FileUpload.js
VD39: This has been done using the type option, if array it will be multiple, else single upload.

src/components/02_atoms/FileUpload/FileUpload.js
VD39: I have updated this to use the typography component that Material supplies.

src/components/02_atoms/FileUpload/FileUpload.js
dawehner: We could fix that later :)

src/components/02_atoms/Widgets/FileUploadWidget.js
```diff
@@ -1,72 +1,199 @@
...
+  bundle,
+  onChange,
+  fieldName,
+  inputProps,
+  entityTypeId,
+  schema: { properties, maxItems },
+}) => {
+  const length = (value && value.data && Object.keys(value.data).length) || 0;
+  const multiple = properties.data.type === 'array'; // If array then allow for multiple uploads
+  const maxItemsCount = maxItems || 1; // maxItems is only set if array, so set to 1 as default
```
dawehner: Isn't the logic? ```multiple ? (maxItems || 100000000000) : 1``` instead?

VD39: Ah, I assumed maxItems was always set if array. Fixed.

Note to self: Don't assume anything. :P

dawehner: Thank you!

Issue #244, opened 2018-07-26
Save a users in-progress node form edits which could be restored later.


lauriii:
As an enhancement to show on the demo we could auto save locally all the modifications user has made on the content edit form. Then when user returns on the page, if there's a saved changes in the local storage, we could prompt the user if they would like to retrieve their previous changes.

lauriii:
How about using snackbars to at least show a message: https://material-ui.com/demos/snackbars/

dawehner:
Nice idea!

dawehner:
Given we want to use redux for this piece, I think this work is blocked on https://github.com/jsdrupal/drupal-admin-ui/issues/251

dawehner:
As first step I think it would be a bit easier to just support restoring the node add forms. The logic when to ask users to restore the node edit forms is a bit tricky, given someone else might have updated the node in the meantime.

I put some code into https://github.com/dawehner/drupal-admin-ui/tree/issue-254 but it doesn't yet cover any localStorage logic, but rather just stores and fetches node forms in redux, and then asks users to restore it.

lauriii:
@dawehner couldn't we as a first step just store the autosave time and compare that to the entity changed time?

dawehner:
@lauriii Sure, well all I'm saying is that you could finish add support and then do the edit support separately, which might be enough for the demo.

lauriii:
Agreed!

dawehner:
Unassigning myself in case someone wants to take over the edit restore logic.

PR #254, opened 2018-07-31
Ask the user to restore their data

dawehner:
Solves: https://github.com/jsdrupal/drupal-admin-ui/issues/244

TODOs:

* [ ] Take into account the ID/UUID
* [ ] Figure out the correct logic when to ask the user for restoring and when not
* [ ] Figure out why some fields (checkboxes / file upload / entity references) don't restore yet


dawehner:
This approach has a good amount of flaws, so let's that deal with that

PR #259, opened 2018-07-31
Fetch a single document.

mattgrill:
## Issue

Fetching all the schema that OpenAPI provides is a bit much. This reduces that.

## Testing instructions

- Make sure `/node/add/recipe` still works.

### DIFF
--- a/src/actions/content.js
+++ b/src/actions/content.js
@@ -10,22 +10,26 @@ import { MESSAGE_ERROR } from '../constants/messages';
 import { setMessage } from './application';
 
 export const SCHEMA_REQUESTED = 'SCHEMA_REQUESTED';
-export const requestSchema = () => ({
+export const requestSchema = ({ entityTypeId, bundle }) => ({
   type: SCHEMA_REQUESTED,
-  payload: {},
+  payload: { entityTypeId, bundle },
 });
 
 export const SCHEMA_LOADED = 'SCHEMA_LOADED';
-function* loadSchema() {
+function* loadSchema(action) {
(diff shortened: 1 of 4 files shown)


src/reducers/content.js
```diff
@@ -47,7 +47,10 @@ export default (state = initialState, action) => {
     case SCHEMA_LOADED: {
       return {
         ...state,
-        schema: action.payload.schema,
+        schema: {
```
dawehner: What should we do with schema loaded beforehand?

dawehner: Is it okay to throw that information away?

src/actions/content.js
```diff
@@ -10,22 +10,26 @@ import { MESSAGE_ERROR } from '../constants/messages';
...
+export const requestSchema = ({ entityTypeId, bundle }) => ({
   type: SCHEMA_REQUESTED,
-  payload: {},
+  payload: { entityTypeId, bundle },
 });
 
 export const SCHEMA_LOADED = 'SCHEMA_LOADED';
-function* loadSchema() {
+function* loadSchema(action) {
+  const { entityTypeId, bundle } = action.payload;
```
dawehner: I thought about whether we want to cache schema for now, but I think till we are aware of all the potential complexities around that, not caching is the most sane approach.

mattgrill: Yes. Ok. So I think this is good?

dawehner: Yes!

mattgrill:
🙏

PR #260, opened 2018-08-01
Automatically construct UISchema

mattgrill:
## Issue

Currently, UISchema is manually created and must be updated with every field we want to support. The information is already available, _although in two places_, from JSON API. Both `field_config` and `entity_form_display` provide us with enough information to construct a usable document and automatically generate the UI.

There is some work in #259 that will benefit from these changes, but they can happen independently. At some point, the bundle schema should be part of these actions/reducers.


dawehner:
Mh I was hoping we have this logic in php

dawehner:
I think we agreed to move this into a PHP endpoint :)

src/components/05_pages/NodeForm/index.js
```diff
@@ -142,17 +149,23 @@ class NodeForm extends React.Component {
...
-                {Object.entries(this.props.uiMetadata)
-                  .map(([fieldName, { widget, inputProps }]) => {
-                    if (Widgets[widget]) {
+        {this.props.schema &&
+          this.props.uiSchema && (
+            <Paper>
+              <div className={styles.container}>
+                <FormControl margin="normal" fullWidth>
+                  {Object.entries(
+                    createUISchema(
```
dawehner: I'm curious should we move this into redux, so the node form isn't aware of it?

src/components/05_pages/NodeForm/index.js
```diff
@@ -142,17 +149,23 @@ class NodeForm extends React.Component {
...
+                      this.props.uiSchema.formDisplaySchema,
+                      this.props.widgets,
+                    ),
+                  )
+                    .map(([fieldName, { widget, inputProps }]) => {
                       // @todo We need to pass along props.
                       // @todo How do we handle cardinality together with jsonapi
                       // making a distinction between single value fields and multi value fields.
-                      const fieldSchema = this.getSchemaInfo(
+                      const entitySchema = this.getSchemaInfo(
```
dawehner: I'm confused about the new name, it is certainly not the schema of the entire entity, but just of one field.

mattgrill: sorry!

src/components/05_pages/NodeForm/index.js
```diff
@@ -37,10 +31,19 @@ class NodeForm extends React.Component {
     entityTypeId: PropTypes.string.isRequired,
     bundle: PropTypes.string.isRequired,
     requestSchema: PropTypes.func.isRequired,
+    requestUiSchema: PropTypes.func.isRequired,
+    uiSchema: PropTypes.oneOfType([
```
dawehner: I opened up a follow for for this: https://github.com/jsdrupal/drupal-admin-ui/issues/279

dawehner:
With this or we use some more fields and render things on the right order

PR #269, opened 2018-08-07
Improve error handling in case of an empty title

lauriii:
## Screenshot

I'm getting following error on the node form:
screen shot 2018-08-07 at 14 21 37

## Testing instructions

I got this error message when accessing node form. Not sure if this needs some other steps to be reproducable given that no one else reported this before.


dawehner:
Ha I ran into the same problem and wondered what I've done wrong. Thank you for fixing!

Issue #266, opened 2018-08-06
Improve usability of the entity reference widget


lauriii:
There's some small improvements that could be made for the entity reference widget:

- [x] The field removes the focus after key down meaning that when searching for the entity, user has to re-focus after every added character.
- [ ] Entities already referenced in the field are still visible in the search results when adding more references.
- [x] Unfiltered list of entities is shown when there's no characters entered and key down is triggered

And here's a gif demonstrating these usability problems:
![entity_reference_widget](https://user-images.githubusercontent.com/1845495/43704816-0c3206f2-998b-11e8-9e95-f75d0797c455.gif)

dawehner:
Thank you for clearly specifying why the behaviour is a bit weird indeed right now.

Do you think this is an issue we could mark as "good first issue"?

martinfrances107:
@lauriii  .. Hi If you think this issue is resolved, would you consider closing it?

We have 60 issues open at the moment... and I am trying to prioritize which one to go for next.

lauriii:
One of the issues reported in this issue wasn't fixed yet. Should we open another issue for that or re-open this?

PR #278, opened 2018-08-07
Improve focus functionality of EntityReferenceAutocomplete

porkloin:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/266

## Screenshot / UI changes

N/A

## Testing instructions

Visit /node/add/recipe – click any of the autocomplete entity reference fields (Recipe category, or Tags) and begin typing. Focus should remain on the text field consistently while typing.


src/components/02_atoms/Widgets/EntityReferenceAutocomplete.js
```diff
@@ -205,7 +205,6 @@ class EntityReferenceAutocomplete extends React.Component {
               fullWidth: true,
               label: this.props.label,
               InputProps: getInputProps({
-                disabled: this.state.loading,
```
mattgrill: There is perhaps a bit of nuance here. If the request is inflight, do we want to send another request on every keystroke?

porkloin: Yeah, that makes a lot of sense. I'll take another stab at it and post another commit soon.

lauriii: >  If the request is inflight, do we want to send another request on every keystroke?

Why wouldn't we send another request on every keystroke?

lauriii:
Currently the suggestions paper is closed after a keystroke. To make the interaction feel smoother, I think that the suggestions should be visible all the time.

entity_reference_widget_modal_flicker

src/components/02_atoms/Widgets/EntityReferenceAutocomplete.js
mattgrill: I think in general it's a bad idea to send a network request on each keystroke. That's potentially hundreds of requests for a single search. Perhaps adding a slight delay to attempt to guess when the user is done typing and issue the request then?

lauriii: Should we consider that if we run into any problems? I suggest that, for now, we do the request on every keystroke and improve it later if that is necessary.

dawehner: Drupal core ran with that strategy for a long time. I think that seems fine for itself

Issue #284, opened 2018-08-08
FileUploadWidget storybook entry has minor problems.


martinfrances107:
#249 Recently adds lots of improvements 

but now when looking at the storybook site I see a few errors injected into the console.

I am just creating a issue before a PR.

martinfrances107:
Here is the PR. I have left one class of error unfixed...

fetches to 'localhost:9001/url' but I am planning on trapping them in a separate issue.

martinfrances107:
Ah crap I made a mistake ... I have created a PR to remove the unwanted dependancy on fetch-mock

martinfrances107:
Thanks for the quick response.

PR #285, opened 2018-08-08
Storybook: Resolve errors displayed on the console.(#284)

martinfrances107:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/284

1) yarn run storybook to start the storybook server
2) Open chrome's devtools view the console.
3) Visit the FileUpload widget storybook definitions

see that most of the errors are gone.

We could trap the fetchs to http://localhost:9001/url but I want to keep this PR simple and deal with that in a follow up.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "vfancy",
   "version": "0.1.0",
-  "license" : "MIT",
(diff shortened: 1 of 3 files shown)


lauriii:
Thank you!

Issue #265, opened 2018-08-05
Create a snackbar message when content got created/updated


dawehner:
Once https://github.com/jsdrupal/drupal-admin-ui/pull/171 is in it would be nice to show a message to the user that the content got updated/created using https://material-ui.com/demos/snackbars/#snackbars

I think we should leverage the text from the existing PHP forms:
```
    if ($insert) {
      $this->logger('content')->notice('@type: added %title.', $context);
      $this->messenger()->addStatus($this->t('@type %title has been created.', $t_args));
    }
    else {
      $this->logger('content')->notice('@type: updated %title.', $context);
      $this->messenger()->addStatus($this->t('@type %title has been updated.', $t_args));
    }
```

porkloin:
I have most of this working, including the following:

- a general purpose snackbar component in the `<Default />` component
- Redux action/reducer for issuing snackbar notifications to the component

However, node save operations don't seem to be working on my local with the a recent code pull (as of 4ish hours ago?) so I'm having some trouble testing issuing a snackbar message on the save operation.

Forgive me if I'm just missing some critical piece of info on how node save is working atm, but once I can get my local working I'd be happy to finish this off!

mattgrill:
I'm pretty sure this is complete now.

PR #289, opened 2018-08-08
General use "Snackbar" component

porkloin:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/265

## Screenshot / UI changes

screen shot 2018-08-08 at 9 03 04 am

## Testing instructions

This is a general-purpose component that can be used for issuing snackbar notifications. The `<SnackbarMessage />` component is mounted at `<Default />` and you can issue a Snackbar notification with `props.summonSnackbar("Snackbar message")`


src/reducers/snackbar.js
```diff
@@ -0,0 +1,28 @@
+import { SUMMON_SNACKBAR, DISMISS_SNACKBAR } from '../actions/snackbar';
+
+export const initialState = {
+  message: '',
+  open: false,
+};
+
+export default (state = initialState, action) => {
+  switch (action.type) {
+    case SUMMON_SNACKBAR: {
```
lauriii: Wondering if there is a better word than summon since it wasn't obvious to me at first without checking a dictionary (I'm a non-native English speaker). How about just OPEN_SNACKBAR?

Maybe a follow-up: we may also want to consider if we want to support setting multiple snackbars simultaneously.

src/actions/snackbar.js
```diff
@@ -0,0 +1,11 @@
+export const SUMMON_SNACKBAR = 'SUMMON_SNACKBAR';
+export const DISMISS_SNACKBAR = 'DISMISS_SNACKBAR';
+
+export const summonSnackbar = message => ({
```
dawehner: I'm wondering whether this should be called snackbar. Snackbar is the "just" the visualization concept, not the underlying idea of showing messages to users. Maybe that's just an OCD thing for me, but maybe using "messages"/"message" in the redux level seems more usable when a different pattern appears.

porkloin: That makes a lot of sense to me – the current terminology kind of shoehorns this into a single-use Redux action/reducer instead of being a generalized interface for issuing messages to the user regardless of the component the message finally renders at. 

With that in mind, I got to grepping around and it looks like we already have setMessage/clearMessage Redux actions that can definitely be utilized to that end with some modifications.

I'm going to throw up another commit here in a sec that refines some of the logic behind the setMessage/clearMessage actions so that they can be more broadly applied.

src/reducers/snackbar.js
porkloin: Great point – a lot of this got swept up into the changes that dawehner mentioned, so the semantics of the messaging should be resolved too. I also made the component work with multiple issued messages!

porkloin:
Alright, so I added another commit which is my second pass at things. I've done the following:

- refactored the setMessage redux action to take `message`, `messageInterface`, and `messageSeverity` as payloads, as well as defaults for `open` (can be used to implement dismissability for banner messages).
- Updated all instances of `<Message>` as `<BannerMessage>` and redux calls to setMessage to use `MESSAGE_INTERFACE_BANNER` to indicate where the message needs to be rendered.
- Updated the `<SnackbarMessage>` component in `<Default>` to render all messages of type `MESSAGE_INTERFACE_BANNER` rather than a single message at a time.

There's still a few todos for me here, namely giving the snackbar notifications an actual way to use the severity level - right now it all renders the same regardless of severity level.

Do we also want to make Banner messages dismissable? With the way the logic of setMessage has been extended it should be pretty straightforward to do.

Lemme know what you guys think!

src/actions/application.js
```diff
@@ -21,21 +24,28 @@ export const SET_MESSAGE = 'SET_MESSAGE';
 
 /**
  *
- * @param {string} message
- * @param type - one of the message constants exported from constants/messages.js
+ * @param {string} message - the message content
+ * @param {string} renderedBy - which of the renderable message areas –
+ *  listed at constants/messages.js
+ * @param {string} severity - the severity level of the message, one of the levels
```
dawehner: Note: the comments seems out of date now

src/actions/snackbar.js
dawehner: Reviewing it again this PR is so much nicer!

src/reducers/application.js
```diff
@@ -43,18 +43,22 @@ export default (state = initialState, action) => {
...
         key: Date.now() + Math.random(),
+        open: true,
       });
       return {
         ...state,
         messages,
       };
     }
     case CLEAR_MESSAGE: {
+      const messages = [...state.messages];
```
dawehner: Should we support clearing all messages somehow?

porkloin: Sure! The next commit (which I just pushed) has that as a feature, and I also removed the now-unnecessary stuff from my original commit (actions/reducers for snackbar).

src/components/02_atoms/InlineMessage/InlineMessage.js
```diff
@@ -0,0 +1,44 @@
...
+  success: css`
+    background-color: #d4edda;
+  `,
+  error: css`
+    background-color: #f8d7da;
+  `,
+};
+
+modifierClassMap = {
+  [MESSAGE_SEVERITY_ERROR]: styles.error,
```
dawehner: Why do we not handle the warning case?

src/components/02_atoms/InlineMessage/InlineMessage.js
```diff
@@ -0,0 +1,44 @@
...
+const Message = ({ message, messageSeverity }) => (
+  <div className={`${styles.message} ${modifierClassMap[messageSeverity]}`}>
+    {message}
+  </div>
+);
+
+Message.propTypes = {
+  message: PropTypes.string.isRequired,
+  messageSeverity: PropTypes.oneOf([
+    MESSAGE_SEVERITY_ERROR,
```
dawehner: I would expect the warning case also be allowed here

Issue #280, opened 2018-08-07
Indicate which fields are required


dawehner:
Some of the fields are meant to be required to have a valid recipe. For example the image.

## Proposal

Using the required flags on the schema set required on https://material-ui.com/api/form-label/#__next

lauriii:
This has been done to fields that are not relationships because of upstream bug: https://www.drupal.org/project/schemata/issues/2991871 so keeping this still open.

PR #290, opened 2018-08-08
Mark fields as required

mattgrill:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/280

## Screenshot / UI changes

create_recipe

## Testing instructions

See if there are `*` on some fields.

### DIFF
--- a/src/components/05_pages/NodeForm/index.js
+++ b/src/components/05_pages/NodeForm/index.js
@@ -179,6 +179,9 @@ class NodeForm extends React.Component {
       onChange: (propType === 'attributes'
         ? this.onAttributeChange
         : this.onRelationshipChange)(fieldName),
+      required: this.props.schema.properties.data.properties.attributes.required.includes(
+        fieldName,
+      ),
       inputProps,
     });
   };
--- a/src/components/05_pages/NodeForm/index.js
+++ b/src/components/05_pages/NodeForm/index.js
@@ -49,6 +49,8 @@ class NodeForm extends React.Component {
     uiSchema: false,
   };
 
+  state = {};
+
   static getDerivedStateFromProps(props, prevState) {
     if (!props.schema) {
       return prevState;
--- a/src/components/05_pages/NodeForm/index.js
+++ b/src/components/05_pages/NodeForm/index.js
@@ -223,16 +223,17 @@ class NodeForm extends React.Component {
 
 styles = {
   fieldContainer: css`
-    padding: 20px 45px 40px;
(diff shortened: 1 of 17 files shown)


src/components/05_pages/NodeForm/index.js
```diff
@@ -189,7 +189,12 @@ class NodeForm extends React.Component {
                         onChange: (propType === 'attributes'
                           ? this.onAttributeChange
                           : this.onRelationshipChange)(fieldName),
-                        inputProps,
+                        inputProps: {
+                          ...inputProps,
+                          required: this.props.schema.properties.data.properties.attributes.required.includes(
```
lauriii: I'm not sure if adding this to input props is a good approach. It seems like this could have a broader effect than just how the input is rendered. Could we just pass a prop required?

dawehner: I think I do agree with lauriii, required seems such a generic thing that it should live on the component itself. Note: We also don't check the propType yet, so required relationships aren't taken into account

mattgrill: I had only attached this to `inputProps` since it seemed sensible since those props should be passed to the `input` element.

dawehner: Don't we actually want to set it on the form control as well, so the label indicates the requiriedness?

src/components/05_pages/NodeForm/index.js
```diff
@@ -189,6 +189,9 @@ class NodeForm extends React.Component {
                         onChange: (propType === 'attributes'
                           ? this.onAttributeChange
                           : this.onRelationshipChange)(fieldName),
+                        required: this.props.schema.properties.data.properties.attributes.required.includes(
```
lauriii: This doesn't take into account yet required relationship fields. However, it seems like schemata don't expose this information yet so I opened an issue to their queue:  https://www.drupal.org/project/schemata/issues/2991871 . Maybe we should just add a todo comment?

src/components/02_atoms/Widgets/TimestampDatetime.js
```diff
@@ -14,6 +14,7 @@ const TimestampDatetime = ({ fieldName, label, value, onChange }) => (
     InputLabelProps={{
       shrink: true,
     }}
+    required={required}
```
lauriii: We should add the newly added prop to `WidgetPropTypes` so that it gets validated.

Issue #282, opened 2018-08-08
Entity reference widget doesn't take field cardinality limitations into account


lauriii:
Currently, the entity reference widget doesn't take into account the field cardinality. The widget should support both; single value and array of values that is limited.

For example, the author field should only allow a single value:
![screen shot 2018-08-08 at 13 30 37](https://user-images.githubusercontent.com/1845495/43820208-c64dd5d6-9b0f-11e8-826e-5cc3bed24f63.png)

VD39:
This issue may now be closed.

dawehner:
Indeed, thank you for your PR!

PR #291, opened 2018-08-08
Entity reference widget doesn't take field cardinality limitations into account

VD39:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/282

## Screenshot / UI changes

Added a check to see if the current widget supports single value.

I couldn't find the max value in the schema, so this PR assumes there are an unlimited number of selection that can be made.

## Testing instructions

Test author (single value) and tags (multiple value).


dawehner:
Seems like a perfect fit. This ensures both that the suggestions aren't requested and potential suggestions returned by some outstanding request, aren't leaked in

dawehner:
> I couldn't find the max value in the schema, so this PR assumes there are an unlimited number of selection that can be made.

Well, it is the same as for the file reference, so maybe we should deal with both cases at the same time

src/components/02_atoms/Widgets/EntityReferenceAutocomplete.js
```diff
@@ -147,13 +151,30 @@ class EntityReferenceAutocomplete extends React.Component {
       );
   };
 
+  cannotAddMoreEntities = () => {
```
lauriii: This is very nitpicky but the name of this function sounds a bit strange to me. How about just changing this to `getMaxItems` and then doing the comparison in the calling code?

VD39: Sure, although the function does two checks, if is single and has an item, or if the max item has been entered. How about changing it to maxItemsSelected?

dawehner: I like @lauriii's idea, given we might want to render some description like: "N items left"

VD39: I have updated the method, I used the same method from the file upload widget to check the maxItems.

Issue #242, opened 2018-07-26
Add support for the secondary column


lauriii:
The Seven uses pattern to have a primary and secondary column depending on the importance of the field. This patter is also included in the Seven re-design work so we know that it will exist in the future. We should implement the secondary column using Material UI for now.

Screenshot of the Seven re-design:
![screen shot 2018-07-26 at 22 13 18](https://user-images.githubusercontent.com/1845495/43271200-20161f04-9121-11e8-9442-f954cf6d841a.png)

PR #292, opened 2018-08-08
Secondary column support

mattgrill:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/242

## Screenshot / UI changes

create_recipe

_LOLOL_

## Testing instructions

- Are there two columns?

_Alright_, this will need some styling help. Additionally since the information about which fields are supposed to exist in the secondary column is not available via an API, the field names are hard coded.

### DIFF
--- a/src/components/02_atoms/FileUpload/FileUpload.js
+++ b/src/components/02_atoms/FileUpload/FileUpload.js
@@ -10,19 +10,13 @@ import api from './../../../utils/api/api';
 const Element = styled('div')`
   border: 2px dashed grey;
   border-radius: 3px;
-  padding: 80px;
-  width: 100%;
-  position: relative;
+  display: flex;
+  padding: 25px;
 `;
 
 const Text = styled('div')`
-  position: absolute;
-  top: 50%;
-  left: 50%;
-  transform: translate(-50%, -50%);
   margin: 0;
   text-align: center;
-
   .icon {
     margin-left: 10px;
   }
--- a/src/components/05_pages/NodeForm/index.js
+++ b/src/components/05_pages/NodeForm/index.js
@@ -3,7 +3,6 @@ import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { css } from 'emotion';
 
-import FormControl from '@material-ui/core/FormControl';
 import Paper from '@material-ui/core/Paper';
 import Divider from '@material-ui/core/Divider';
 import Button from '@material-ui/core/Button';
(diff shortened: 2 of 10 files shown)


src/components/05_pages/NodeForm/index.js
```diff
@@ -145,66 +145,66 @@ class NodeForm extends React.Component {
     schema.properties.data.properties.attributes.properties[fieldName] ||
     schema.properties.data.properties.relationships.properties[fieldName];
 
+  createField = ({ fieldName, widget, inputProps }) => {
```
dawehner: Super nice to extract that into its own helper function

src/components/05_pages/NodeForm/index.js
```diff
@@ -145,66 +145,66 @@ class NodeForm extends React.Component {
...
+  };
+
   render() {
+    let result = null;
+    if (this.props.schema && this.props.uiSchema) {
+      const { right, left } = createUISchema(
+        this.props.uiSchema.fieldSchema,
+        this.props.uiSchema.formDisplaySchema,
+        this.props.widgets,
+        ['promote', 'status', 'sticky'],
```
dawehner: I think hardcoding the second column is totally a pragmatic choice right now! 👍

src/utils/api/schema.js
```diff
@@ -39,33 +44,42 @@ const createUISchema = (fieldSchema, formDisplaySchema, widgets) =>
...
+            ? fieldSchema[currentFieldName].attributes.settings
+            : {}),
+          ...(Object.prototype.hasOwnProperty.call(
+            formDisplaySchema,
+            currentFieldName,
+          )
+            ? formDisplaySchema[currentFieldName].settings
+            : {}),
+        };
+        acc[
```
dawehner: I'm wondering whether we should keep the existing createUiSchema function and instead have a second function which creates the left/right information out of an existing schema. ❓

src/components/02_atoms/FileUpload/FileUpload.js
```diff
@@ -10,19 +10,13 @@ import api from './../../../utils/api/api';
 const Element = styled('div')`
   border: 2px dashed grey;
   border-radius: 3px;
-  padding: 80px;
-  width: 100%;
-  position: relative;
+  display: flex;
```
dawehner: Does someone know the state of flexbox IE 11 support?

src/components/05_pages/NodeForm/index.js
```diff
@@ -145,78 +148,91 @@ class NodeForm extends React.Component {
...
+        ? this.onAttributeChange
+        : this.onRelationshipChange)(fieldName),
+      inputProps,
+    });
+  };
+
   render() {
+    let result = null;
+    if (this.props.schema && this.props.uiSchema) {
+      const { right, left } = sortUISchemaFields(
```
dawehner: Thank you!

src/utils/api/schema.js
```diff
@@ -68,4 +68,15 @@ const createUISchema = (fieldSchema, formDisplaySchema, widgets) =>
       return acc;
     }, []);
 
-export { createEntity, createUISchema };
+const sortUISchemaFields = (schema, secondaryColumnFields) =>
+  schema.reduce(
```
dawehner: Nice use of reduce!

src/components/02_atoms/FileUpload/FileUpload.js
martinfrances107: https://caniuse.com/#feat=flexbox

Under the known issues tab - there is a good definition of "partial" support.

PR #297, opened 2018-08-09
204 response support

lauriii:
## Issue

It seems like after https://www.drupal.org/project/jsonapi/issues/2977653 204 is also used by JSON API. Add support for that.


lauriii:
I was thinking of that but didn't do it because we cannot handle all 2xx request yet (for example 205) so I though it would be better to add them later when they are needed.

dawehner:
Good point.

Also, interesting, I just learned about 205 :)

PR #299, opened 2018-08-09
Use ckeditor to render basic HTML textfields

dawehner:
image


dawehner:
Sadly this didn't work, let's see us in https://github.com/jsdrupal/drupal-admin-ui/pull/323

Issue #304, opened 2018-08-09
Bulk operations not disabled after submit


mortenson:
Currently at `/admin/content`, bulk operations ("Actions") are disabled until content is selected. Once you select content and select an action, I expected the bulk operations to go back to a disabled state but they're still available, which leads to errors.

PR #305, opened 2018-08-09
Make sure the bulk operations state is reset correctly.

mattgrill:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/304

## Screenshot / UI changes

state

## Testing instructions

- Ensure bulk operations are not allowed after none are selected.

### DIFF
--- a/src/components/05_pages/Content/Content.js
+++ b/src/components/05_pages/Content/Content.js
@@ -147,6 +147,7 @@ class Content extends Component {
       action => action.attributes.id === this.state.action,
     )[0];
     this.props.actionExecute(matchingAction, Object.keys(this.state.checked));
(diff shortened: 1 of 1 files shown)


src/components/05_pages/Content/Content.js
```diff
@@ -147,6 +147,7 @@ class Content extends Component {
       action => action.attributes.id === this.state.action,
     )[0];
     this.props.actionExecute(matchingAction, Object.keys(this.state.checked));
+    this.setState({ checked: {}, action: false });
```
mortenson: `action` defaults to `null` - should we set it back to that or is `false` OK?

mattgrill: 👍 Good point.

Issue #301, opened 2018-08-09
FileUploadWidget: Keep Storybook in sync with component changes


martinfrances107:
The latest commit #297 broke the storybook entry for FileUploadWidget ( multiple ) 

No biggie, just want to keep these things in sync.

I am looking at this now...

martinfrances107:
I am about to crate a PR 

Here is the main misconception, I have unpicked 

The fileItem data structure which is passed into the FileUploadWidget

it can be a file data structure or an array of many fileItem data structures 

but the change is to recognise the fileItem data structure must carry alternative text on a per item basis.

```
<FileUploadWidget value={data: [ 
      {file:{/*data struct */}, meta: {/* data struct */}}
      {file:{/*data struct */}, meta: {/* data struct */}}
      ...
]} /> 
```

Previously the code changes allowed for only one alternative  meta text string.

Anyway I hope this description is useful in showing the problem ..

PR is coming

dawehner:
There are basically two ways the value could look like:

### Cardinality 1
```
{{"type":"file--file","id":"0b538928-09de-438d-9f33-15e0f2c104a1","meta":{"alt":"A delicious deep layered Mediterranean quiche with basil garnish.","title":null,"width":768,"height":511}
```

### Cardinality > 1

```
[{"type":"file--file","id":"426b7e5d-19f4-423c-a40d-355cb910d5fa","meta":{"alt":"Fresh cut herbs including mint, parsley, thyme and dill","title":"","width":768,"height":512}},{"type":"file--file","id":"302a93f7-dedd-40ad-b030-31be5bdb4317","meta":{"alt":"test","title":"","width":2494,"height":1416}}]}}
```

martinfrances107:
Thanks for helping me out with this .... it is much appreciated.

Is the data structure you have provided based on the ideal .. what is expected in core ?
Is is what should be pushed to the server as a json object.?

As a mini experiment I added this debug  so I can see the item being filtered 

here is my code snippet. [ NodeForm:render()]

```
  const items = getItemsAsArray(multiple, value.data)
    // Default schema creates stub entries, which we don't need here.
    .filter(item => {
      console.log(item);
      return item.id;
    });
  console.log('filtered');
  console.log(items);
```

Visiting the "add recipe" page and uploading  a jpg here is the data structure I see.

id: xyz
file: { id: xyz .... }
meta: { alt:'' }

so id is repeated .. and as the code stands meta is outside the file object .. not inside!

[...]

PR #316, opened 2018-08-11
Fix bug in getItemsAsArray().

martinfrances107:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/301

## Screenshot / UI changes

The visual appearance of the storybook entries for the FileUploadWidget are restored to their former glory.

UI changes ... yes

FileUploadWidget: the property value has had its data structures revamped to allow for alternate meta string on a per file basis ( During multiple uploads ).

## Testing instructions

run yarn storybook and see:

the errors are removed.
the FileUploadWidget entries are restored.

### DIFF
--- a/src/components/02_atoms/Widgets/FileUploadWidget.js
+++ b/src/components/02_atoms/Widgets/FileUploadWidget.js
@@ -60,7 +60,7 @@ const FileUploadWidget = ({
 
   const items = getItemsAsArray(multiple, value.data)
     // Default schema creates stub entries, which we don't need here.
-    .filter(item => item.file.id);
+    .filter(item => item.id);
   const length = (items && items.length) || 0;
   // maxItems is only set if array, so set to 1 as default.
   const maxItemsCount = multiple ? maxItems || 100000000000 : 1;
@@ -112,12 +112,12 @@ const FileUploadWidget = ({
             <Card>
               <CardContent>
                 <List>
-                  {items.map((data, index) => {
+                  {items.map((item, index) => {
                     const {
-                      file,
+                      id,
                       meta: { alt },
-                    } = data;
-                    const { id, url, filename } = file;
+                      file: { url, filename },
+                    } = item;
                     const last = items.length - 1 === index;
 
                     return (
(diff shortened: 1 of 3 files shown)


martinfrances107:
My fix is on the right lines... but significantly now on the recipe page the FileUploadWidget is broken 

I will to correct this ....in the morning.

src/components/02_atoms/Widgets/FileUploadWidget.js
```diff
@@ -60,7 +60,7 @@ const FileUploadWidget = ({
 
   const items = getItemsAsArray(multiple, value.data)
     // Default schema creates stub entries, which we don't need here.
-    .filter(item => item.id);
+    .filter(item => item.file.id);
```
dawehner: Are you sure about this change? There should be an id

martinfrances107: Sure ? No.. but I think maybe yes and we need to adjust our code.

 I am still looking into .. What I am trying to understand/correct is :

If you look at the storybook definition ( showing the Widget with multiple files already 'added' )

items is an array of objects with the following structure { file: {} meta{} } 

Currently when the FileUploadWidget is constructed as part of the recipe form just the 'file' object is passed in as item(s)

It makes sense to me that with say 3 files uploaded there are 3 distinct alternate text string to process/display

What I need to take care of an get right is how 'file' and 'meta' are iterated over and transported through the render function and how the defaults are initialized/derived from the schema.

<FileUploadWidget value={data: [ 
      {file:{/*data struct */}, meta: {/* data struct */}}
      {file:{/*data struct */}, meta: {/* data struct */}}
      ...
]} /> 

Anyway I hope this is a better explanation of my current mental block.

martinfrances107:
For review here is a high level list of changes

1) There was a bug in getItemsAsArray
2) For the storybook pages the correct files structure is now passes into the FileUploadWidget

data: {
  {
    id: XYZ
    file: {}
    meta: {}
  },
  {
    id: XYZ
    file: {}
    meta: {}
  },
... }
3) I have made the individual data items unique. [ a new id number per file uploaded].

Issue #319, opened 2018-08-13
NodeForm: warns about initialization.


martinfrances107:
When I visit 

http://localhost:3000/node/add/recipe

I see this warning in the Chrome's devtool console output.

> Warning: NodeForm: Did not properly initialize state during construction. Expected state to be an object, but it was undefined.

We are relying on the default constructor, and have a getDerivedStateFromProps() method.

I think we need either a static state variable, or a explicit constructor 

I see a separate small blemish, in NodeForm

`static getDerivedStateFromProps(props, prevState) 
`
 makes use of props.entity -- props a function input not the class variable  this.props

Formally  entity is a implied, undeclared property. -- this code is a work in progress
and this.props.entity is a variable which is written to but never used.

I will try and make this better.

dawehner:
This is resolved now, 🎆

PR #320, opened 2018-08-13
NodeForm initialisation generates warnings(#319)

martinfrances107:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/319

Visit 

http://localhost:3000/node/add/recipe

open Chrome's dev tools .. look at the console log..

before the PR there was a NodeForm: initialization warning

after the PR the warning is resolved.

### DIFF
--- a/src/components/05_pages/NodeForm/index.js
+++ b/src/components/05_pages/NodeForm/index.js
@@ -46,6 +46,11 @@ class NodeForm extends React.Component {
     uiSchema: false,
   };
 
+  constructor(props) {
+    super(props);
+    this.state = {};
+  }
+
   static getDerivedStateFromProps(props, prevState) {
     if (!props.schema) {
       return prevState;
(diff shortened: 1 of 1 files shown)


martinfrances107:
> I think we can just do state = {}; and avoid the constructor function.

This does not work unfortunately.

I think the problem  is some poorly documented interaction when a class has a getDerivedStateFromProps() method.

This is what fails.

```
 static defaultProps = {
    schema: false,
    uiSchema: false,
  };

  static state = {};

  static getDerivedStateFromProps(props, prevState) {
    if (!props.schema) {
      return prevState;
    }
```

mattgrill:
So I did just test this, 

state

mattgrill:
Oh, I also see what happened. `static state = {}` is probably not what you want. CRA is probably using, https://babeljs.io/docs/en/next/babel-plugin-proposal-class-properties.html underneath.

martinfrances107:
Oh dear god..  what was I thinking.

I haven't made a mistake that big since I wandered around nursery school without any trousers on!

PR #323, opened 2018-08-14
Initial draft.js based editor

dawehner:
Note: I went with draft.js as it was less work ;)

screen shot 2018-08-14 at 01 05 25

### DIFF
--- a/src/components/02_atoms/Widgets/TextTextarea.css
+++ b/src/components/02_atoms/Widgets/TextTextarea.css
@@ -1,1 +1,0 @@
-.public-DraftEditor-content { min-height: 110px; }
--- a/src/components/02_atoms/Widgets/TextTextarea.css
+++ b/src/components/02_atoms/Widgets/TextTextarea.css
@@ -1,0 +1,1 @@
+.public-DraftEditor-content { min-height: 110px; }
--- a/src/components/02_atoms/Widgets/TextTextarea.js
+++ b/src/components/02_atoms/Widgets/TextTextarea.js
@@ -1,8 +1,10 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
+import { css } from 'emotion';
 import RichTextEditor from 'react-rte';
-import './TextTextarea.css';
 import WidgetPropTypes from '../../05_pages/NodeForm/WidgetPropTypes';
+
+let styles;
 
 /**
  * Some basic wysiwyg editor based upon react-rte which is based upon draft.js.
@@ -42,9 +44,23 @@ class TextTextarea extends Component {
   };
 
   render() {
-    return <RichTextEditor value={this.state.value} onChange={this.onChange} />;
+    return (
+      <RichTextEditor
+        className={styles.container}
+        value={this.state.value}
+        onChange={this.onChange}
+      />
+    );
   }
 }
+
+styles = {
+  container: css`
+    .public-DraftEditor-content {
+      min-height: 110px;
+    }
+  `,
+};
 
 TextTextarea.defaultProps = {
   value: {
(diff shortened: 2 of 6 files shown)


src/components/02_atoms/Widgets/TextTextarea.css
```diff
@@ -0,0 +1 @@
+.public-DraftEditor-content { min-height: 110px; }
```
justafish: Can you move this to emotion? I don't think we're using external css files anywhere else (it's supported because it came with CRA)

dawehner: I'm not sure how I can access the nested component in there

mattgrill: It looks like this implementation already uses css files (the need for a css loader in webpack), so there's no js->css api 🤔

src/components/02_atoms/Widgets/TextTextarea.js
```diff
@@ -42,10 +44,24 @@ class TextTextarea extends Component {
...
+      <RichTextEditor
+        className={styles.container}
+        value={this.state.value}
+        onChange={this.onChange}
+      />
+    );
   }
 }
 
+styles = {
```
dawehner: Nice, thank you!

src/components/02_atoms/Widgets/TextTextarea.js
```diff
@@ -0,0 +1,72 @@
...
+ */
+class TextTextarea extends Component {
+  static propTypes = {
+    ...WidgetPropTypes,
+    value: PropTypes.shape({
+      value: PropTypes.string.isRequired,
+    }),
+  };
+
+  constructor(props) {
```
mattgrill: I don't think we need this constructor function, and can just use `state = {}`

src/components/02_atoms/Widgets/TextTextarea.js
```diff
@@ -0,0 +1,72 @@
...
+      value: RichTextEditor.createValueFromString(
+        (this.props.value && this.props.value.value) || '',
+        'html',
+      ),
+    };
+  }
+
+  onChange = value => {
+    this.setState({ value });
+    if (this.props.onChange) {
```
mattgrill: Should we make sure this happens in the callback of `setState`?

src/components/02_atoms/Widgets/TextTextarea.js
```diff
@@ -0,0 +1,72 @@
...
+
+styles = {
+  container: css`
+    .public-DraftEditor-content {
+      min-height: 110px;
+    }
+  `,
+};
+
+TextTextarea.defaultProps = {
```
mattgrill: Maybe put these in the `Component`?

src/components/02_atoms/Widgets/TextTextarea.js
```diff
@@ -0,0 +1,72 @@
...
+
+let styles;
+
+/**
+ * Some basic wysiwyg editor based upon react-rte which is based upon draft.js.
+ * This was designed to be as easiest as possible for the demo.
+ *
+ * On the longrun we might want to switch back to ckeditor.
+ */
+class TextTextarea extends Component {
```
mattgrill: I think we use `React.Component` in other places?

dawehner: 👍

src/components/02_atoms/Widgets/TextTextarea.js
dawehner: 👍

src/components/02_atoms/Widgets/TextTextarea.js
dawehner: 👍

Issue #251, opened 2018-07-30
Allow editing existing content


dawehner:
https://github.com/jsdrupal/drupal-admin-ui/pull/171 would enable us to save new content

Once that is in, let's try to see whether we can get editing working as well

mattgrill:
Alright, #171 has been merged. This is unblocked now.

dawehner:
🎆

mattgrill:
* [ ] Posting fails right now because it sends the status field which is not allowed to be sent due to workflow moderation. Ideally the PATCH would just POST changed fields
* [ ] Ensure saving actually works

lauriii:
https://github.com/jsdrupal/drupal-admin-ui/pull/379 fixes this 🎉

PR #328, opened 2018-08-14
Enabling editing a node

dawehner:
https://github.com/jsdrupal/drupal-admin-ui/issues/251

Current TODOs:

* [x] Figure out how to query for all referenced entities when fetching the content
* [x] Fix the file upload to include the referenced file
* [x] Ensure all content actually gets prepoulated: categories, authors, tags, images, authored by are all empty right now
* [ ] Posting fails right now because it sends the status field which is not allowed to be sent due to workflow moderation. Ideally the PATCH would just POST changed fields
* [ ] Ensure saving actually works

### DIFF
--- a/src/actions/content.js
+++ b/src/actions/content.js
@@ -115,14 +115,17 @@ export const requestSingleContent = (bundle, id) => ({
 
 export const CONTENT_SINGLE_LOADED = 'CONTENT_SINGLE_LOADED';
 function* loadSingleContent(action) {
-  const { payload: { bundle, id } } = action;
+  const {
+    payload: { bundle, id },
+  } = action;
   try {
     yield put(resetLoading());
     yield put(showLoading());
 
-    const content = yield call(api, 'content_single', { parameters: { bundle, id }, queryString: {
-
-      }});
+    const content = yield call(api, 'content_single', {
+      parameters: { bundle, id },
+      queryString: {},
+    });
     yield put({
       type: CONTENT_SINGLE_LOADED,
       payload: {
--- a/src/actions/content.js
+++ b/src/actions/content.js
@@ -107,6 +107,35 @@ function* loadContent(action) {
   }
 }
 
+export const CONTENT_SINGLE_REQUESTED = 'CONTENT_SINGLE_REQUESTED';
+export const requestSingleContent = (bundle, id) => ({
+  type: CONTENT_SINGLE_REQUESTED,
+  payload: { bundle, id },
+});
+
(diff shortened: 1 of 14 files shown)


src/routes.js
```diff
@@ -15,7 +16,15 @@ const routes = {
   '/admin/reports/dblog': Dblog,
   '/node/add': AddContent,
   '/node/add/recipe': () => (
-    <NodeForm entityTypeId="node" bundle="recipe" widgets={widgets} />
+    <NodeAddForm entityTypeId="node" bundle="recipe" widgets={widgets} />
+  ),
+  '/node/:bundle/:id/edit': ({ match }) => (
```
dawehner: Note: We need to pass along the bundle as there is not way to load a single piece of content without knowing the bundle.

src/actions/content.js
```diff
@@ -107,6 +107,35 @@ function* loadContent(action) {
...
+});
+
+export const CONTENT_SINGLE_LOADED = 'CONTENT_SINGLE_LOADED';
+function* loadSingleContent(action) {
+  const { payload: { bundle, id } } = action;
+  try {
+    yield put(resetLoading());
+    yield put(showLoading());
+
+    const content = yield call(api, 'content_single', { parameters: { bundle, id }, queryString: {
```
dawehner: Should we first look up in state.content.contentById whether the content already exists?

src/components/02_atoms/Widgets/FileUploadWidget.js
```diff
@@ -50,153 +51,220 @@ const styles = {
...
+      } else {
+        ids = [this.props.value.data.id];
+      }
+      this.fetchEntitites(entityTypeId, bundle, ids).then(({ data: items }) => {
+        this.setState({
+          selectedItems: items.map(({ id, attributes }, index) => ({
+            id,
+            type: 'file--file',
+            [entityTypeId]: attributes,
+            meta: this.props.value.data[index].meta,
```
mattgrill: @dawehner Here's an interesting issue. If there is only one entity returned, you can't fetch it by it's index; it's at `this.props.value.data.meta`. You can probably check like, `Array.isArray(this.props.value.data) ? this.props.value.data[index].meta : this.props.value.data.meta`

dawehner: good point!

mattgrill:
Alright. 🎉👍 Great job! I'm really excited to get this in, and give us some more time to address the issues in a series of followup issues.

Issue #274, opened 2018-08-07
Show human readable label on OptionsSelect widget


lauriii:
Currently the `OptionsSelect` widget is showing the machine name of the options on the UI. It would be a nice enhancement to show the human readable name in the UI ✨

<img width="158" alt="screen shot 2018-08-07 at 18 21 59" src="https://user-images.githubusercontent.com/1845495/43773104-ce2c5bca-9a6e-11e8-8eeb-5bd6e3d5051f.png">

<img width="202" alt="screen shot 2018-08-07 at 18 22 27" src="https://user-images.githubusercontent.com/1845495/43773125-dc88a6d8-9a6e-11e8-94e9-26bd21230d6a.png">

mattgrill:
Alright. Had a quick look at this. It doesn't actually look like the human readable names are returned from `schemata`. Additionally, both `entity_form_display` and `field_config` do not return the human readable names.

wimleers:
Which human readable name is not being returned by `schemata`? The label of each allowed value in the allowed value list?

Also, why is `schemata` relevant here? Wouldn't this be something one would get by reading the `field_storage_config` config entity?

---

I just created a `field_options` field on `article` `node`s with the same allowed values list as in the screenshot in the OP. `http://d8/jsonapi/field_storage_config/field_storage_config/017f2e79-ceff-472e-9c5a-606c134facc1` returns:
```
{
  "data": {
    "type": "field_storage_config--field_storage_config",
    "id": "017f2e79-ceff-472e-9c5a-606c134facc1",
    "attributes": {
      "langcode": "en",
      "status": true,
      "dependencies": {
        "module": [
          "node",
          "options"
        ]
      },
      "drupal_internal__id": "node.field_op",
      "field_name": "field_op",
      "entity_type": "node",
      "field_storage_config_type": "list_string",
      "settings": {
        "allowed_values": {
          "easy": "Easy",
          "medium": "Medium",
          "hard": "Hard"
        },
        "allowed_values_function": ""
      },
      "module": "options",
      "locked": false,
      "cardinality": 1,
      "translatable": true,
      "indexes": [],
      "persist_with_no_fields": false,
      "custom_storage": false
    },
    "links": {
      "self": "http://d8/jsonapi/field_storage_config/field_storage_config/017f2e79-ceff-472e-9c5a-606c134facc1"
    }
  },
  "jsonapi": {
    "version": "1.0",
    "meta": {
      "links": {
[...]

PR #336, opened 2018-08-15
Use human readable names on Options Select widget.

mattgrill:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/274

## Screenshot / UI changes

create_recipe

## Testing instructions

- Human readable names should be available on the options-select widget.

### DIFF
--- a/src/actions/schema.js
+++ b/src/actions/schema.js
@@ -37,7 +37,7 @@ function* loadUiSchema(action) {
           filter: { targetEntityType: entityTypeId, bundle, mode: 'default' },
         },
       }),
-      api('field_storage', {
+      api('field_storage_config', {
         queryString: {
           filter: { condition: { path: 'entity_type', value: entityTypeId } },
         },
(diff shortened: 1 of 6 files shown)


src/actions/schema.js
```diff
@@ -36,6 +37,7 @@ function* loadUiSchema(action) {
           filter: { targetEntityType: entityTypeId, bundle, mode: 'default' },
         },
       }),
+      api('field_storage'),
```
dawehner: May I ask one question: It would be nice to just load the field storage entries we need. it should be possible to restrict the fields per entity type and bundle

src/components/02_atoms/Widgets/OptionsSelect.js
```diff
@@ -20,8 +20,9 @@ const OptionsSelect = ({
   helpText,
   fieldName,
   required,
-  schema: { enum: options, default: defaultValue },
```
dawehner: Let's adopt the proptypes for ```OptionsSelect``` to document the new requirements we have

src/utils/api/api.js
```diff
@@ -158,6 +158,10 @@ async function api(
       url = '/jsonapi/entity_form_display';
       break;
     }
+    case 'field_storage': {
```
dawehner: Why do we call it ```field_storage``` and not ```field_storage_config``` here?

Issue #333, opened 2018-08-15
FileUploadWidget: Invalid prop definition for multiple uploads


martinfrances107:
Regarding Chrome console log : -

a)  localhost:9000/add/recipe
b) Also the two storybook sections for the FileUploadWidget

> Warning: Failed prop type: Invalid prop `value.data` of value `[object Object]` supplied to `FileUploadWidget`, expected one of [null,null].

This is a sister issue to  

#331 which moves a fullWidth property up into a FormControl

#325 Which will fix a  recurring key issue.

martinfrances107:
I have changed the title because this is now effectively a follow up to #337 .

It fixes issues that remain for the case of multiple files uploads.

dawehner:
The current list of proptypes seem to not longer come from the ```FileUploadWidget```. Given that I'll close this :)

PR #332, opened 2018-08-15
(#331) BooleanCheckbox add FormControl.

martinfrances107:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/331

We need a FormControl element if we are going to use the fullWidth mechanism.

## Testing instructions

Visit localhost:3000/add/recipe

open Chrome's devtools look at the console log.

With the PR the fullWidth warning as described in the main issue is resolved.

Parsing of the FormControl can now continue and gives more meaningful messages messages about the repeating warning ( which makes  #325  easier to debug/review)

Issue #331, opened 2018-08-15
Wrap BooleanCheckBox in a FormControl element


martinfrances107:
I am splitting off a small side issue from 

#325 Second Column warning 

This warning is not like the others

React does not recognize the `fullWidth` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `fullwidth` instead. If you accidentally passed it from a parent component, remove it from the DOM element.

fullWidth is a FormControl concept and is invalid on a FormControlLabel or a CheckBox element.

PR is on its way.

PR #337, opened 2018-08-15
Cleanup console errors due to bad props and PropType checking.

mattgrill:
## Screenshot / UI changes

create_recipe

## Testing instructions

- no more console errors.

### DIFF
--- a/src/components/02_atoms/Widgets/FileUploadWidget.js
+++ b/src/components/02_atoms/Widgets/FileUploadWidget.js
@@ -209,7 +209,10 @@ const fileItemSchema = PropTypes.shape({
 FileUploadWidget.propTypes = {
   ...WidgetPropTypes,
   value: PropTypes.shape({
-    data: PropTypes.oneOf([PropTypes.arrayOf(fileItemSchema), fileItemSchema]),
+    data: PropTypes.oneOfType([
+      PropTypes.arrayOf(fileItemSchema),
+      fileItemSchema,
+    ]),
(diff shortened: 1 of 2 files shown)


src/components/02_atoms/Widgets/FileUploadWidget.js
```diff
@@ -209,7 +209,10 @@ const fileItemSchema = PropTypes.shape({
 FileUploadWidget.propTypes = {
   ...WidgetPropTypes,
   value: PropTypes.shape({
-    data: PropTypes.oneOf([PropTypes.arrayOf(fileItemSchema), fileItemSchema]),
+    data: PropTypes.oneOfType([
```
dawehner: Oh, you are absolute right!

martinfrances107:
Ah I was working in parallel on these issues.


But here are the difference between your solution and mine 

Regarding the deletion of fullWidth my solution in #332 , #331  was to move the fullWidth up into a FormControl parent/wrapper... I am not sure which I prefer.

As far as I can see yes this patch removes the errors on the /add/recipe page.

But the solution for the case of both single and multiple FileUploadWidget Storybook defintions
takes a little more work see (#338) and (#333)



Anyway I hope I have spoken with a neutral tone..

mattgrill:
@martinfrances107 Sure! :) I dropped the `fullWidth` prop on `CheckBox` as it's not supported and gets passed incorrectly to the DOM. If you want to add it to the parent in another PR, feel free. I think it would be great. :)

Issue #321, opened 2018-08-13
FileUploadWidget: Adjust semantic structure 


martinfrances107:
In this issue I want to looking at FileUploadWidget and is styling .... after we have dragged images into the active area.

Things are currently working .. but some of the semantic/glue structure is a little awkward.

We make use of material-ui's Card element... with has a set of associated child elements 

```
<Card>
  <CardHeader >
  <CardMedia ... />
  <CardContent>
    ...
  </CardContent>
  <CardAction>
  ... buttons etc
  <CardActions>
</Card>
```

I want to use this card template -  

1) Give every uploaded file its own card. 
2) Replace the Image element inside the card and use what is expected a CardMedia element.
3) Nest the associated delete button inside the CardActions section.

A) This will make it easier for developers down the line style to the  widget in anyway they desire. 
B) Improve accessibility.

The following image is what the styling looks like when every uploaded element gets its own card. A card with a separated actions section.

I am about the push a PR 

![cardthumbnail](https://user-images.githubusercontent.com/1259969/44058340-c5555d04-9f45-11e8-823a-a899c033b987.png)

dawehner:
@martinfrances107 @ckrina send me this monosnap: https://monosnap.com/file/Mjd3GBGZ5F2GFvYHIq1OoCCFgX8oJO which is not necessarily complete, but I think keeping the files where they are make sense

PR #338, opened 2018-08-15
FileUploadWidget: Invalid prop definition for multiple uploads (#333)

martinfrances107:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/333

## Testing instructions

This reduces one of the many  warnings that appear on the console log when visiting 

for http://localhost:3000/node/add/recipe
[ The warnings were also visible when looking at the storybook definitions of the FileUploadWidget ]

The parent issue has a description of the bug. But here is a short summary of the solution.

There were two bugs one hiding behind the other

The first bug was a misconception... 

PropTypes.oneOf() incorrectly instead of PropTypes.oneOfType() 

The second bug was just to get the PropTypes to make sense

the id which is required is different in the two distinct cases.

### DIFF
--- a/src/components/02_atoms/Widgets/FileUploadWidget.js
+++ b/src/components/02_atoms/Widgets/FileUploadWidget.js
@@ -262,7 +262,17 @@ class FileUploadWidget extends React.Component {
   }
 }
 
-const fileItemSchema = PropTypes.shape({
+const fileItemMultiplePropType = PropTypes.shape({
+  id: PropTypes.toString.isRequired,
+  file: PropTypes.shape({
+    type: PropTypes.string.isRequired,
+    url: PropTypes.string.isRequired,
+    id: PropTypes.string,
+    filename: PropTypes.string.isRequired,
+  }),
+});
+
+const fileItemPropType = PropTypes.shape({
   file: PropTypes.shape({
     type: PropTypes.string.isRequired,
     url: PropTypes.string.isRequired,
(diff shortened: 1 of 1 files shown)


dawehner:
Thank you for fixing these prop types, I wish we would have a nicer language :)

Sorry there is a conflict now :(

martinfrances107:
Ah no problem.. there was a flurry of commits within about 10mins .... but hey it is good to see problems go away. I will fix this up in the morning.

martinfrances107:
Ok so this is now a follow-up issue

.. so I have rebased on the new master and forced a push...

It removes the error... but now FileUploadWidget can be parsed a little more, revealing a new set of error messages. These warnings are related to the bad 'key" structure of the List .. ListItem structure 

I will solve those bugs as part of 

#321 - FileUploadWidget: Adjust the semantic structure of

src/components/02_atoms/Widgets/FileUploadWidget.js
```diff
@@ -262,23 +262,36 @@ class FileUploadWidget extends React.Component {
   }
 }
 
-const fileItemSchema = PropTypes.shape({
+const fileItemMultiplePropType = PropTypes.shape({
+  id: PropTypes.toString.isRequired,
+  file: PropTypes.shape({
```
dawehner: Should we at least share the file part of this proptype?

martinfrances107: Thanks, that is a good idea.

PR #343, opened 2018-08-16
Make it possible to POST articles

dawehner:


lauriii:
Does this break saving recipe entities?

Issue #342, opened 2018-08-16
/node/add - ul elements need a set of li children


martinfrances107:
Screenreaders are very sensitive to the DOM structure of ol, ul elements 

Failure to maintain  a ul li parent-child relationship breaks accessibility.

At the moment we only have a set of A elements as the direct children of a UL element 

The images below shows that the axe report for the /node/add page.

It needs a little TLC.

![axeulissue](https://user-images.githubusercontent.com/1259969/44206825-fa16f980-a152-11e8-8054-dd0de415a5a9.png)
.

I have fixed a similar issue so I know what the problem is. It should be a one line fix.. I am working on this now.

martinfrances107:
Thank you, for the rapid response as always. 💯

PR #344, opened 2018-08-16
List a11y issue

martinfrances107:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/342

## Testing instructions

as per instruction on the home page enable aXe with : -

REACT_APP_AXE=true yarn start

visit http://localhost:3000/node/add/recipe

in the parent issue I detailed a axe report that highlights a issue with the DOM structure which confounds screen readers.

With this PR applied the aXe warnings go away.

Please note a small amount of css is needed to bypass a underline text issue common to many  LINK elements.

### DIFF
--- a/src/components/05_pages/AddContent/AddContent.js
+++ b/src/components/05_pages/AddContent/AddContent.js
@@ -11,6 +11,10 @@ import ListItemText from '@material-ui/core/ListItemText';
 import PageTitle from '../../02_atoms/PageTitle';
 
 const styles = {
+  menuLink: css`
+    display: inherit;
+    text-decoration: inherit;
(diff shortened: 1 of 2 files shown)


dawehner:
Oh wow, that's unexpected from material UI!

martinfrances107:
The nighwatch was failing 'cos I did not account for the insertion of a extra li layer. A Simple fix.

martinfrances107:
>Oh wow, that's unexpected from material UI!

If you look on stack overflow the old pattern for adding links into lists is frequently quoted as a
as a good thing todo ... it is a shame that it breaks screenreaders.

PR #345, opened 2018-08-16
Make filters visible even if there's no results

lauriii:

### DIFF
--- a/src/components/05_pages/Content/Content.js
+++ b/src/components/05_pages/Content/Content.js
@@ -193,8 +193,7 @@ class Content extends Component {
               <Fragment>
                 <div className={styles.filters}>
                   {this.props.contentTypes &&
-                    this.props.actions &&
-                    this.props.contentList.length >= 1 && (
+                    this.props.actions && (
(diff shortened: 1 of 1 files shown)


src/components/05_pages/Content/Content.js
```diff
@@ -206,7 +205,6 @@ class Content extends Component {
                             });
                           }}
                           margin="normal"
-                          disabled={!this.props.contentList.length || false}
```
mattgrill: I had specifically disabled these to prevent interaction if there was no content to be filtered. I don't think it makes sense to allow a user to filter 'nothing'.

lauriii: The problem is that you might select a set of filters that is too restrictive and doesn't find any results - and after that, the filters would be disabled meaning that only way to move back to less restrictive filter set is to refresh the page 😅

mattgrill: Makes sense!

dawehner: Oh great point!

PR #346, opened 2018-08-16
Improve content filter styles

lauriii:
## Screenshot / UI changes
screen shot 2018-08-16 at 21 29 21


lauriii:
@mattgrill I think it's easier to find when it's on its own line. I was on a call with Sally and were both confused "where's the actions?" for a good while.

PR #347, opened 2018-08-16
Execute all actions when multiple actions are spawned at once

lauriii:


lauriii:
@mattgrill Currently, when you select multiple articles to delete (using bulk actions), it only deletes one of them. If you have suggestions how this could be solved in another way, I'm open to any ideas! 🙂

dawehner:
```From looking at the documentation takeLatest appears to cancel a previous saga if one is still running when a new action is dispatched. However takeEvery will spawn a new saga regardless if a previous one is still running.
```
Yeah, it makes sense for things which don't have parameters / just appear once, like loading the menu content, schema etc.

PR #348, opened 2018-08-16
Use Drupal route pattern for edits

mattgrill:
## Testing instructions

- Make sure `/node/:nid/edit` works


src/utils/api/api.js
```diff
@@ -58,7 +58,7 @@ async function api(
       options.headers.Accept = 'application/vnd.api+json';
       break;
     case 'content_single':
-      url = `/jsonapi/node/${parameters.bundle}/${parameters.id}`;
```
dawehner: How about not touching the content_single endpoint but keep using the content one instead, given it is now identical?

src/components/05_pages/NodeEditForm/index.js
```diff
@@ -32,21 +38,18 @@ NodeEditForm.defaultProps = {
...
-  bundle: PropTypes.string.isRequired,
-  id: PropTypes.string.isRequired,
+  nid: PropTypes.string.isRequired,
   requestSingleContent: PropTypes.func.isRequired,
   entity: PropTypes.shape({
-    data: PropTypes.shape({
-      attributes: PropTypes.shape({
-        title: PropTypes.string.isRequired,
-      }).isRequired,
+    attributes: PropTypes.shape({
```
dawehner: It feels like a follow up could expand the prop type a bit more to include relationships as well?

src/components/05_pages/NodeEditForm/index.js
```diff
@@ -8,20 +8,26 @@ import PageTitle from '../../02_atoms/PageTitle/PageTitle';
...
 
   render() {
-    const { entity, bundle } = this.props;
+    const { entity } = this.props;
     return (
       <Fragment>
         {entity && (
           <PageTitle>
-            Edit {bundle} {entity.data.attributes.title}
+            Edit {entity.type.replace('node--', '')} {entity.attributes.title}
```
dawehner: Should we add the bundle as variable here?

src/components/05_pages/NodeEditForm/index.js
mattgrill: Yes. I want to create a followup to address PropTypes across all our components.

src/utils/api/api.js
mattgrill: Yes. Good point!

PR #351, opened 2018-08-17
Make node title a link

lauriii:
## Issue

screen shot 2018-08-17 at 20 08 09


martinfrances107:
I am happy with this change....but there is subtle loss of A11y functionality - that I just want to make public.

Screen readers are still happy as the select input has a label..

but users with reduced fine motor control now have a much smaller target to select if they toggle the checkbox.

Anyway I don't think it is a problem.

PR #353, opened 2018-08-17
Move created and uid fields to the right column on node form

lauriii:
## Screenshot / UI changes

screen shot 2018-08-17 at 20 59 04


mattgrill:
At some point we should move these fields to another file, but for now it's great.

PR #357, opened 2018-08-18
First iteration of adding support for restoring entity forms

dawehner:
Note: This doesn't use localStorage yet, given it is not strictly needed for the demo


src/components/05_pages/NodeForm/index.js
```diff
@@ -41,44 +42,18 @@ class NodeForm extends React.Component {
...
 
   static defaultProps = {
     schema: false,
     uiSchema: false,
+    onChange: () => {},
   };
 
-  state = {};
-
-  static getDerivedStateFromProps(props, prevState) {
```
dawehner: I moved the state calculation to ```componentDidMount/componentDidUpdate``` due to https://reactjs.org/docs/react-component.html#componentdidupdate especially because ```getDerivedStateFromProps``` causes an infinite loop

src/components/05_pages/NodeForm/index.js
```diff
@@ -41,44 +42,18 @@ class NodeForm extends React.Component {
       }),
       PropTypes.bool,
     ]),
+    onChange: PropTypes.func,
   };
 
   static defaultProps = {
     schema: false,
     uiSchema: false,
+    onChange: () => {},
```
dawehner: Note: Due to storing content in redux typing seemed to be a little bit slow. Maybe that's just me?

src/components/05_pages/NodeForm/index.js
```diff
@@ -179,9 +199,41 @@ class NodeForm extends React.Component {
     });
   };
 
+  renderRestoreSnackbar() {
+    return (
+      // Ensure that there was some previously stored entity
```
dawehner: It feels like this logic isn't right yet ...

src/reducers/content.js
```diff
@@ -19,6 +21,14 @@ export default (state = initialState, action) => {
           [action.payload.content.data.id]: action.payload.content,
         },
       };
+    case CONTENT_ADD_CHANGE:
```
dawehner: For restoring node edit forms we should store it by entity ID/UUID instead of by bundle. That's the reason there is a dedicated action for adds, but sure this could be unified

src/actions/content.js
```diff
@@ -150,6 +150,12 @@ function* loadSingleContent(action) {
   }
 }
 
+export const CONTENT_ADD_CHANGE = 'CONTENT_ADD_CHANGE';
```
mattgrill: Just from quickly taking a look, I'm not sure this name feels correct. I'm not totally sure what is happening just from this action. Not sure if that makes sense?

src/components/05_pages/NodeForm/index.js
```diff
@@ -1,11 +1,12 @@
-import React from 'react';
+import React, { Fragment } from 'react';
```
mattgrill: We added `Fragment` here, but I don't see it being used?

mattgrill:
Also looks like there are some conflicts with the recent changes.

src/actions/content.js
dawehner: Yeah, so action is for storing the state of the actual node add form

src/components/05_pages/NodeForm/index.js
dawehner: Damnit I pushed to the wrong remote

src/actions/content.js
dawehner: Do you have a better suggestion? This is basically about storing node form content in redux

dawehner:
@lauriii Thank you for testing, I've found a way to fix that
@fpersico Not sure, what would you expect as a user?

dawehner:
I fixed a good amount of problems. @mattgrill's comment is still up, maybe someone has a better name suggestion.

dawehner:
I'm quite sure there are still issues when doing some manual testing

src/components/05_pages/NodeForm/index.js
```diff
@@ -132,25 +109,76 @@ class NodeForm extends React.Component {
...
+      }),
+      () => this.props.onChange(this.props.bundle, this.state.entity),
+    );
   };
 
   getSchemaInfo = (schema, fieldName) =>
     schema.properties.data.properties.attributes.properties[fieldName] ||
     schema.properties.data.properties.relationships.properties[fieldName];
 
+  calculateState = (prevProps, prevState, setState) => {
```
lauriii: I'm wondering if there would be a better name for this function; maybe `refreshState` would be more accurate?

lauriii: Or `rebuildState`?

src/components/02_atoms/Widgets/TextTextarea.js
```diff
@@ -37,25 +37,49 @@ class TextTextarea extends React.Component {
...
+      (Array.isArray(props.value) &&
+        props.value.length &&
+        props.value[0].value) ||
+        props.value.value ||
         '',
       'html',
-    ),
+    );
+
+  extractValueString = props =>
```
lauriii: 🤯

src/components/05_pages/NodeForm/index.js
dawehner: How about ```deriveState```?

Issue #356, opened 2018-08-18
Reducer type error. 


martinfrances107:
Error when visiting

 http://localhost:3000/node/add

as seen in the Chrome's console logs

index.js:2178 

uncaught at loadContentTypes TypeError: action.payload.contentTypes.data.reduce is not a function
    at ./src/reducers/application.js.__webpack_exports__.a

francescopersico:
I can't reproduce this bug. No console error on http://localhost:3000/node/add.

francescopersico:
Got the cause of the problem. Going to fix it.

PR #363, opened 2018-08-20
Reducer type error fixed.

vishalchandna1:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/356

## Screenshot / UI changes

Added the validation to check the Object instance.

## Testing instructions
1. Go to homepage `/`
2. Click on the `+` button at the bottom right corner.
3. There should be no error in the console window.

### DIFF
--- a/src/reducers/application.js
+++ b/src/reducers/application.js
@@ -126,18 +126,22 @@ export default (state = initialState, action) => {
       };
     }
     case CONTENT_TYPES_LOADED: {
+      const { data } = action.payload.contentTypes;
       return {
         ...state,
-        contentTypes: action.payload.contentTypes.data.reduce(
-          (accumulator, contentType) => ({
-            ...accumulator,
-            [contentType.attributes.type]: {
-              name: contentType.attributes.name,
-              description: contentType.attributes.description,
-            },
-          }),
-          {},
-        ),
+        contentTypes:
+          data instanceof Array
+            ? data.reduce(
+                (accumulator, contentType) => ({
+                  ...accumulator,
+                  [contentType.attributes.type]: {
+                    name: contentType.attributes.name,
+                    description: contentType.attributes.description,
+                  },
+                }),
+                {},
+              )
+            : data,
       };
     }
     case ACTIONS_LOADED: {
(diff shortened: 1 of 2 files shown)


src/reducers/application.js
```diff
@@ -126,18 +126,22 @@ export default (state = initialState, action) => {
...
-          (accumulator, contentType) => ({
-            ...accumulator,
-            [contentType.attributes.type]: {
-              name: contentType.attributes.name,
-              description: contentType.attributes.description,
-            },
-          }),
-          {},
-        ),
+        contentTypes:
```
dawehner: Can we add the comment how this data could be empty? Is this a backend returning nothing?

src/reducers/application.js
```diff
@@ -126,18 +126,22 @@ export default (state = initialState, action) => {
...
+                (accumulator, contentType) => ({
+                  ...accumulator,
+                  [contentType.attributes.type]: {
+                    name: contentType.attributes.name,
+                    description: contentType.attributes.description,
+                  },
+                }),
+                {},
+              )
+            : data,
```
justafish: let's return an empty object instead of `data` as the fallback here

mattgrill: @justafish I think what will happen, _although I haven't tested it_, is that if you only have one Content Type, instead of an `array`, its just an object.

mattgrill: Which is an especially frustrating part of the API.

vishalchandna1:
@mattgrill  A good catch. I will remove the cache lookup part and update the PR. Thanks for your suggestions 👍

vishalchandna1:
@mattgrill  @justafish  Removed the cache lookup and updated the PR. Please review.

src/reducers/application.js
lauriii: After the changes made for `loadContentTypes`, wouldn't we be able to revert all changes made to this file?

vishalchandna1: Hi @lauriii  I think it is better to have validations in place to counter any unaccounted error.

mattgrill: @vishalchandna1 Looks like these changes aren't actually needed anymore. The response `data` will never not be an array so having this here will just confuse people in the future.

vishalchandna1:
Hi @mattgrill  , okay I have reverted that code now.

lauriii:
@vishalchandna1 thanks, looks good!

Issue #293, opened 2018-08-09
Populate uid (author) field automatically


lauriii:
The current UI auto-populates the uid (author) field. We should implement this for our client as well.

<img width="646" alt="screen shot 2018-08-09 at 14 29 33" src="https://user-images.githubusercontent.com/1845495/43884578-ac76b4b2-9be0-11e8-92ae-0ba0e01b8233.png">
<img width="365" alt="screen shot 2018-08-09 at 14 29 53" src="https://user-images.githubusercontent.com/1845495/43884587-b2f61562-9be0-11e8-9410-fe4adecf075d.png">

dawehner:
Nice idea!

lauriii:
This now works in the scope of the demo, but we should extend this to work with other users as uid 1.

mattgrill:
Lets revisit this once we migrate to JSON API 2.x which should get us the actual user.

mattgrill:
@martinfrances107 JSON API 2.x has had a release, however we are not depending on it in the support module.

PR #367, opened 2018-08-20
Set a default for authored by field.

mattgrill:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/293

## Screenshot / UI changes

user

## Testing instructions

- Make sure there is a default user for new content.

### DIFF
--- a/src/components/05_pages/NodeForm/index.js
+++ b/src/components/05_pages/NodeForm/index.js
@@ -80,6 +80,9 @@ class NodeForm extends React.Component {
       )
       .reduce((agg, [key, value]) => ({ ...agg, [key]: value }), {});
 
+    // @TODO Remove this when we create the entity scaffolding at the NodeEdit
+    // or NodeAdd component level.
+    // https://github.com/jsdrupal/drupal-admin-ui/issues/378
     // Set default `Authored By` relationship.
     if (!Object.prototype.hasOwnProperty.call(props, 'entity')) {
       state.entity.data.relationships.uid.data = { ...props.user };
--- a/src/components/05_pages/NodeForm/index.js
+++ b/src/components/05_pages/NodeForm/index.js
@@ -9,7 +9,7 @@ import Button from '@material-ui/core/Button';
 
 import SchemaPropType from './SchemaPropType';
 
-import { contentAdd } from '../../../actions/content';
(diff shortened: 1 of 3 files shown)


src/components/05_pages/NodeForm/index.js
```diff
@@ -77,10 +79,16 @@ class NodeForm extends React.Component {
           .includes(key),
       )
       .reduce((agg, [key, value]) => ({ ...agg, [key]: value }), {});
+
+    // Set default `Authored By` relationship.
+    if (!Object.prototype.hasOwnProperty.call(props, 'entity')) {
```
lauriii: Did you see if this could be added in the `NodeAddForm` component? It would be nice if it's added there so it's clear that it is only applied on the entity creation.

src/components/05_pages/NodeForm/index.js
```diff
@@ -77,10 +79,16 @@ class NodeForm extends React.Component {
...
+      state.entity.data.relationships.uid.data = { ...props.user };
+    }
     return state;
   }
 
   componentDidMount() {
+    this.props.requestUser(1);
```
lauriii: Shouldn't this be the current user rather than user 1?

lauriii: It seems like you can just make a GET request to `/jsonapi/user` which returns the current user 👻

lauriii: Sorry, I overlooked that a bit since I wasn't logged in as uid 1. However, it seems like there's https://www.drupal.org/project/jsonapi/issues/2927037 which has been fixed in 2.x.

mattgrill: Yeah. So we should probably consider upgrading to 2.x at some point.

src/components/05_pages/NodeForm/index.js
mattgrill: Done.

src/components/05_pages/NodeForm/index.js
mattgrill: Here's a followup to upgrade to 2.x, https://github.com/jsdrupal/drupal-admin-ui/issues/380

PR #369, opened 2018-08-20
Save edits using PATCH.

mattgrill:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/251

## Testing instructions

- save edits to an existing node
- make a new one.


lauriii:
Any thoughts if this could use https://github.com/jsdrupal/drupal-admin-ui/pull/373?

dawehner:
By looking at #373 and this PR it feels like a better approach would be:

1) Merge #373 
2) Introduce a content_edit action / api which encapsulates the HTTP logic

PR #372, opened 2018-08-21
Add missing message severity on add content error

lauriii:

### DIFF
--- a/src/actions/content.js
+++ b/src/actions/content.js
@@ -314,7 +314,7 @@ function* addContent({ payload: { content } }) {
     yield put(showLoading());
     yield call(api, 'node:add', { parameters: { node: content } });
   } catch (error) {
-    yield put(setMessage(error.toString()));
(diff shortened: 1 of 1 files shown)


src/actions/content.js
```diff
@@ -314,7 +314,7 @@ function* addContent({ payload: { content } }) {
     yield put(showLoading());
     yield call(api, 'node:add', { parameters: { node: content } });
   } catch (error) {
-    yield put(setMessage(error.toString()));
+    yield put(setMessage(error.toString(), MESSAGE_SEVERITY_ERROR));
```
dawehner: Just in general, I'm wondering: Could we just introduce a ```setErrorMessage``` which makes this way easier to read?

lauriii: Sounds like a good suggestion to me!

lauriii: Created new issue for this: https://github.com/jsdrupal/drupal-admin-ui/issues/375

PR #373, opened 2018-08-21
Make onSave specific to the action entity form is tied for

lauriii:
Move onSave from the NodeForm to it's parent so that changing it separately for entity creation and editing is possible.

### DIFF
--- a/src/components/05_pages/NodeAddForm/index.js
+++ b/src/components/05_pages/NodeAddForm/index.js
@@ -1,19 +1,43 @@
 import React, { Fragment } from 'react';
+import { connect } from 'react-redux';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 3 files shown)


src/components/05_pages/NodeAddForm/index.js
```diff
@@ -1,19 +1,43 @@
...
-  <Fragment>
-    <PageTitle>Create {props.bundle}</PageTitle>
-    <LoadingBar />
-    <NodeForm {...props} />
-  </Fragment>
-);
+class NodeAddForm extends React.Component {
+  constructor() {
+    super();
+    this.onSave = this.onSave.bind(this);
```
dawehner: Note: If you use ```onSave = (entity) => {``` instead, you don't have to bind the function

PR #376, opened 2018-08-21
Make type check in PageTitle component more accurate

lauriii:
## Issue

Sometimes the document title wasn't set because of an overly strict type check. This should improve the situation and make the title visible.


lauriii:
For anyone wondering; the behavior of the instanceof was very surprising to me. According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof#Examples simple strings will return false on `instanceof String`.

Issue #377, opened 2018-08-21
Separate out the api.js utils hard-coded strings to a separate JS constant file.


vishalchandna1:
Remove the hard-coded strings from utils/api/api.js with constants.

vishalchandna1:
I am taking this up.

PR #388, opened 2018-08-22
Hard-coded strings separated to a new constant file.

vishalchandna1:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/377

Hard coded strings separated to a constant file for api.js

### DIFF
--- a/src/utils/api/api.js
+++ b/src/utils/api/api.js
@@ -1,5 +1,6 @@
 import qs from 'qs';
 import { ApiError } from './errors';
+import * as apiConstants from '../../constants/api';
 
 async function api(
   endpoint,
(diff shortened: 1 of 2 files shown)


vishalchandna1:
It needs some more work.

justafish:
Thanks for your PR!

I'm not in favour of this change for two reasons:
- It makes it much harder to read/follow/make changes to
- It's a temporary file

Would love to get some other opinions though

lauriii:
@vishalchandna1 Would be great to hear some reasoning why this is important 🙂

src/utils/api/api.js
```diff
@@ -1,5 +1,6 @@
 import qs from 'qs';
 import { ApiError } from './errors';
+import * as apiConstants from '../../constants/api';
```
mattgrill: I agree with @justafish that this makes everything further down much harder to parse.

Some of these strings are only used once, so moving them to a 'constants' file does not really make sense. Things that are used over and over, like `_format=json` or `application/vnd.api+json` perhaps make sense, but those could also just be stored in this file.

dawehner:
For me at least repeating simple code isn't really the problem, but rather repeating complex code.

I could see there is a point in introducing constants for the API types, but to be honest, a better approach would be to introduce generic entity api integrations, so there is no need for any of the repeatitiveness in the first place. @vishalchandna1 Does that sound like an okayish solution for the problem you describe?

justafish:
Thanks for the feedback everyone. Given the others' comments, I'm going to close this PR.

@vishalchandna1 We'll revisit this after Drupal Europe when we look at cleaning up our code after the demo

vishalchandna1:
ok, CoOL! Thanks for your feedbacks :)

Issue #389, opened 2018-08-22
Ensure that snackbars shown one by one


lauriii:
[According to Material UI style guide](https://material-ui.com/demos/snackbars/), snackbars can be only shown one at a time. Currently, if multiple messages are dispatched simultaneously, they are all shown at once. Change the logic that only one snackbar is shown at once.

![screen shot 2018-08-22 at 16 03 40](https://user-images.githubusercontent.com/1845495/44454259-3c45ac80-a625-11e8-9917-a6560b31d5ad.png)

abhaykumar1415:
i want to work on this issue.. thanks

PR #392, opened 2018-08-22
Synchonous snackbar error messages.

vishalchandna1:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/389


src/reducers/application.js
```diff
@@ -55,7 +55,10 @@ export default (state = initialState, action) => {
     }
     case CLEAR_MESSAGE: {
       const messages = [...state.messages];
-      messages.pop(messages.map(message => message.key === action.payload.key));
+      messages.splice(
```
lauriii: Why is this change necessary? 🤔

mattgrill: @lauriii The second parameter for `.splice()` is `deleteCount`. Previously it was omitted, meaning that all elements from the start to the end of the array were removed. Setting it to `1` indicates that we only want to remove one element.

mattgrill: Also this is a switch from `.pop()` to `.splice()`.

dawehner: How about replacing all this code by ```return { ...state, messages: state.messages.filter(message => message.key !== action.payload.key)}```?

vishalchandna1: The main motive to use splice() here is to **remove** an element from the state messages. Using a filter() method will also do the job, I believe.

Let me know if any changes are required :)

Issue #395, opened 2018-08-22
Add validation for required fields


lauriii:
We show information about required fields in the UI but we don't actually validate them. Add some validation to required fields so that posting entities without required fields filled isn't possible. This should take into account the required fields on the node, but also on the relationships. For example, alt text is required when uploading images.

mattgrill:
I think this is complete since both the linked issues are fixed.

PR #396, opened 2018-08-22
Show an alert when you have missing fields

mattgrill:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/395

### DIFF
--- a/src/components/05_pages/NodeForm/index.js
+++ b/src/components/05_pages/NodeForm/index.js
@@ -16,8 +16,7 @@ import MultipleFields from '../../02_atoms/MultipleFields/MultipleFields';
 
 import { contentAdd, requestUser } from '../../../actions/content';
 import { requestSchema, requestUiSchema } from '../../../actions/schema';
-import { setMessage } from '../../../actions/application';
-import { MESSAGE_SEVERITY_ERROR } from '../../../constants/messages';
+import { setErrorMessage } from '../../../actions/application';
 
 import {
   createEntity,
@@ -52,7 +51,7 @@ class NodeForm extends React.Component {
       data: PropTypes.object,
     }),
     requestUser: PropTypes.func.isRequired,
-    setMessage: PropTypes.func.isRequired,
+    setErrorMessage: PropTypes.func.isRequired,
     onChange: PropTypes.func,
   };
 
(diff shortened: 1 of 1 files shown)


src/components/05_pages/NodeForm/index.js
```diff
@@ -147,6 +159,29 @@ class NodeForm extends React.Component {
     schema.properties.data.properties.attributes.properties[fieldName] ||
     schema.properties.data.properties.relationships.properties[fieldName];
 
+  resolveRequiredFiles = () => {
```
fpersico: Files -> Fields

fpersico: Also seem the method name must be something like resolveMissingFields not resolveRequiredFields

src/components/05_pages/NodeForm/index.js
```diff
@@ -147,6 +159,29 @@ class NodeForm extends React.Component {
...
+    const unavailableFields = ['nid', 'uuid', 'vid', 'path'];
+    const requiredFields = this.props.schema.properties.data.properties.attributes.required.filter(
```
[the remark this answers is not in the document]
lauriii: Yeah, we can't take them into account yet because https://github.com/jsdrupal/drupal-admin-ui/issues/280 is not finished yet.

src/components/05_pages/NodeForm/index.js
```diff
@@ -151,6 +163,29 @@ class NodeForm extends React.Component {
...
+      field => !unavailableFields.includes(field),
+    );
+    return Object.entries(this.state.entity.data.attributes)
+      .filter(([fieldName]) => requiredFields.includes(fieldName))
+      .filter(([fieldName, value]) => {
+        if (
+          typeof value === 'object' &&
```
vishalchandna1: Do we need array validations here? The current condition will be true when the value is an Array

src/components/05_pages/NodeForm/index.js
```diff
@@ -121,7 +125,15 @@ class NodeForm extends React.Component {
   };
 
   onSave = () => {
-    this.props.onSave(this.state.entity.data);
+    const missingFields = this.resolveMissingRequiredFields();
+    if (missingFields.length) {
+      this.props.setMessage(
```
dawehner: I think we should have a singular and non singular message here. ```title field_recipe_instruction field_summary are missing.
``` sounds really weird. On top of that, can we extract the title out of the schema?

Comment (dawehner) [reply src/components/05_pages/NodeForm/index.js]: How about using setErrorMessage?

File: src/components/05_pages/NodeForm/index.js [hunk 5]
```diff
@@ -100,7 +103,17 @@ class NodeForm extends React.Component {
...
+        missingFields.length > 1
+          ? `The following fields are missing, ${missingFields.join(', ')}.`
```
dawehner: Nice!

mattgrill: This felt easier that trying to pluralize some bits of the string.

dawehner: Yeah, never ever try to split up text to some bits of the string. This makes it much harder later to introduce localization of the UI.

Issue #394, opened 2018-08-22
Populate created timestamp automatically


lauriii:
The authored on timestamp is currently by default empty. The current UI autogenerates this value and I think we should do the same here.

<img width="913" alt="screen shot 2018-08-23 at 00 15 05" src="https://user-images.githubusercontent.com/1845495/44479394-b2b8cd80-a66a-11e8-9fad-525fcd8288a7.png">
<img width="395" alt="screen shot 2018-08-23 at 00 22 28" src="https://user-images.githubusercontent.com/1845495/44479402-b8161800-a66a-11e8-907d-2038113db930.png">

PR #397, opened 2018-08-22
Set default created on date.

mattgrill:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/394

### DIFF
--- a/src/components/02_atoms/Widgets/DatetimeTimestamp.js
+++ b/src/components/02_atoms/Widgets/DatetimeTimestamp.js
@@ -16,9 +16,9 @@ const DatetimeTimestamp = ({
     fullWidth
     label={label}
     type="datetime-local"
-    defaultValue={value}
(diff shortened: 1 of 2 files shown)


src/components/02_atoms/Widgets/DatetimeTimestamp.js
```diff
@@ -16,9 +16,9 @@ const DatetimeTimestamp = ({
     fullWidth
     label={label}
     type="datetime-local"
-    defaultValue={value}
+    defaultValue={new Date(value * 1000).toJSON().slice(0, 19)}
```
dawehner: May I ask a question: Why do we not provide a default value from the schema?

lauriii: The value is coming from state. JavaScipt uses millisecond timestamp rather than second based used more commonly (which we also use for state) meaning that we have to do some conversions when using JavaScript APIs.

Issue #403, opened 2018-08-23
Show confirmation dialog before bulk deleting content


lauriii:
We have the confirmation dialog already working when deleting a single piece of content on the content list. However, when deleting multiple pieces of content, there's no confirmation which is a bit awkward.

dawehner:
This got fixed, thank you @mattgrill

PR #405, opened 2018-08-23
Show a dialog for bulk operations.

mattgrill:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/403

## Screenshot / UI changes

bulk-actions

### DIFF
--- a/src/components/05_pages/Content/ConfirmationDialog.js
+++ b/src/components/05_pages/Content/ConfirmationDialog.js
@@ -36,9 +36,9 @@ const ConfirmationDialog = ({
       onClose={handleClose}
     >
       <DialogTitle>
-        Are you sure you want to{' '}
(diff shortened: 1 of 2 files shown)


src/components/05_pages/Content/ConfirmationDialog.js
```diff
@@ -0,0 +1,85 @@
...
+    .map(({ attributes: { label } }) => label)
+    .shift();
+  return (
+    <Dialog
+      open={dialogVisibility}
+      TransitionComponent={props => <Slide direction="up" {...props} />}
+      onClose={handleClose}
+    >
+      <DialogTitle>
+        Are you sure you want to{' '}
```
dawehner: The text is a little bit confusing right now: ```Are you sure you want to make content sticky these content items?```. I'm wondering whether we could use something like "Are you sure you want to apply the \"make content sticky\" action on these content items?"```

Issue #195, opened 2018-07-10
Add a multi-delta wrapper to add the INGREDIENTS fields


dawehner:
So far every field was just allowed to provide a single value. The ingredients field though can hold multiple values at once, each being a text field.

* [ ] Implement some wrapper, which allows to enter multiple data
![image](https://user-images.githubusercontent.com/29678/42526496-e1056db6-846d-11e8-975f-8287e362b504.png)
* [ ] Based upon the cardinality of the schema add the wrapper automatically to each field
* [ ] Add the ingredients field

dawehner:
@lhuria94 I think this would work fine for a entity reference widget, but I struggling seeing how this would work for normal textfields?

Drupal has some form of strange requirements

* Some widgets handle multiple files at once (autocomplete widget, file upload)
* Some widgets don't want to handle multiple, like numbers, checkboxes, textfields etc.

It would be nice to come up with some form of abstraction which helps the second case

lhuria94:
This example is with reference to text fields, not just for autocomplete. Like you type some text and enter, similarly, we can add multiple.

Something like this: https://material.angularjs.org/latest/demo/chips

But I see your point, guess this needs more thinking.

VD39:
I will work on this.

PR #410, opened 2018-08-24
195 add a multi delta wrapper

VD39:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/195

## Screenshot / UI changes

- Can add multiple values for inputs.
- Widget and node form changes.
- Design using material UI list.
- Simple sorting using drag and drop.
- Allow only the icon be be draggable and not the whole list item.
- Auto focus when adding new item.

Still TODO:

- Prevent the dragging from outside container.
- Remove ghosting/transparency and display whole element.

I will continue to work on the above, however, this pull request can be merged if okay as is.

## Testing instructions

Test the ingredients, check that adding, deleting, and deleting works.


src/components/02_atoms/MultipleFields/MultipleFields.js
```diff
@@ -0,0 +1,206 @@
...
+
+  /**
+   * Initial state
+   */
+  state = {
+    currentIndex: -1,
+    selectedValue: null,
+  };
+
+  componentDidMount() {
```
dawehner: Note: I think we need a componentDidUpdate for the [restore content functionality](https://github.com/jsdrupal/drupal-admin-ui/issues/244).

src/components/02_atoms/MultipleFields/MultipleFields.js
```diff
@@ -0,0 +1,206 @@
...
+    const {
+      props: { value },
+      state: { currentIndex, selectedValue },
+    } = this;
+
+    const overIndex = parseInt(event.currentTarget.dataset.key, 10);
+    if (currentIndex !== overIndex) {
+      // Changes the two values with one another
+      value[currentIndex] = value[overIndex];
+      value[overIndex] = selectedValue;
```
dawehner: Why do we have to store the selectedValue here? I would have expected that instead of selectedValue you could write 
```
selectedValue = value[currentIndex];
value[currentIndex] = value[overIndex];
value[overIndex] = selectedValue;
```

src/components/02_atoms/MultipleFields/MultipleFields.js
dawehner: I just tried it, and it indeed doesn't work. Do you mind having a look?

src/components/02_atoms/MultipleFields/MultipleFields.js
```diff
@@ -0,0 +1,206 @@
...
+    newValue[index] = value;
+    onChange(newValue);
+  };
+
+  /**
+   * Adds another empty string to the current set of values.
+   */
+  addAnotherItem = () => {
+    const { value, onChange } = this.props;
+    const newValue = [...value, ''];
```
dawehner: I'm curious, what are you thoughts of supporting other kind of widgets, like for example numbers, selected fields etc.?

src/components/02_atoms/MultipleFields/MultipleFields.js
```diff
@@ -0,0 +1,206 @@
...
+                  </Button>
+                </ListItemSecondaryAction>
+              </ListItem>
+            ))}
+          <Add>
+            <Button
+              color="primary"
+              variant="contained"
+              onClick={addAnotherItem}
+              aria-label="Add another item"
```
dawehner: Nice!

src/components/02_atoms/MultipleFields/MultipleFields.js
VD39: Does the data actually restore? I get the notification, however, when I click okay nothing happens to any of the fields.

src/components/02_atoms/MultipleFields/MultipleFields.js
VD39: A good point, fixed.

src/components/02_atoms/MultipleFields/MultipleFields.js
VD39: From what I can see, if it is a input type, it will be an array of string. Maybe we should do the same check as we do when creating the value for single and multiple fields: https://github.com/jsdrupal/drupal-admin-ui/blob/master/src/utils/api/schema.js#L1.

dawehner:
Let's address the rest on a follow up. Amazing work!

Issue #414, opened 2018-08-26
Setting content author doesn't work


francescopersico:
Author field seem not working

1. Go to http://localhost:3000/node/add/article
2. Remove 'admin' as author
3. Than try to add another author
4. The new author get removed from the input field

![author](https://user-images.githubusercontent.com/15610456/44632703-99bc5f00-a97f-11e8-9ff8-7f42d609f9ed.gif)

PR #415, opened 2018-08-26
Fix author (EntityReferenceAutocomplete) field

francescopersico:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/414

### DIFF
--- a/src/utils/api/fieldItem.js
+++ b/src/utils/api/fieldItem.js
@@ -11,6 +11,9 @@ export const setItemById = (multiple, item, items) => {
 };
 
 export const getItemsAsArray = (multiple, items) => {
+  if (Array.isArray(items)) {
+    return items;
+  }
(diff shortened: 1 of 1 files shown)


src/utils/api/fieldItem.js
```diff
@@ -11,6 +11,9 @@ export const setItemById = (multiple, item, items) => {
 };
 
 export const getItemsAsArray = (multiple, items) => {
+  if (Array.isArray(items)) {
```
mattgrill: Do we care if this is an empty array?

dawehner: I think the fix is a proper fix. As follow up I opened up https://github.com/jsdrupal/drupal-admin-ui/issues/420 to ensure we have unit test coverage for those low level functionality.

Issue #418, opened 2018-08-27
EntityReferenceAutocomplete: Add Autocomplete attribute 


vishalchandna1:
Add `autoComplete` attribute in EntityReferenceAutocomplete.js to control the HTML5 browser built in autocomplete feature.

PR #419, opened 2018-08-27
EntityReferenceAutocomplete: Autocomplete attribute added

vishalchandna1:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/418

## Screenshot / UI changes
screen shot 2018-08-27 at 4 42 27 pm

### DIFF
--- a/src/components/02_atoms/Widgets/EntityReferenceAutocomplete.js
+++ b/src/components/02_atoms/Widgets/EntityReferenceAutocomplete.js
@@ -276,10 +276,12 @@ class EntityReferenceAutocomplete extends React.Component {
     <TextField
       label={label}
       id={InputProps.id}
-      // @todo Disable the browser built in autocompletion.
       InputProps={{
         inputRef: ref,
         ...InputProps,
+        inputProps: {
+          autoComplete: 'off',
+        },
       }}
       {...other}
     />


src/components/02_atoms/Widgets/EntityReferenceAutocomplete.js
```diff
@@ -272,14 +272,16 @@ class EntityReferenceAutocomplete extends React.Component {
     );
   };
 
-  renderInput = ({ InputProps, ref, label, ...other }) => (
+  renderInput = ({ InputProps, ref, label, autoComplete, ...other }) => (
```
dawehner: Do you think there is ever a usecase where we want the browser based autosuggestion? I would argue that we don't have to make this configurable. 🤔

vishalchandna1: Hi @dawehner 
I was also not so sure about it, so I added a generic case having `autoComplete` default value as `off`

Should we always keep `autocomplete` as `off`?

dawehner: Yeah, I think the autocompletion by the browser is always annoying 👍

src/components/02_atoms/Widgets/EntityReferenceAutocomplete.js
```diff
@@ -13,7 +13,7 @@ import FormControl from '@material-ui/core/FormControl';
 import WidgetPropTypes from '../../05_pages/NodeForm/WidgetPropTypes';
 import SchemaPropType from '../../05_pages/NodeForm/SchemaPropType';
 
-import api from './../../../utils/api/api';
+import api from '../../../utils/api/api';
```
mattgrill: Is there a reason we need to fix this here?

vishalchandna1: This is just to keep a consistent code like the other `import` statements.

src/components/02_atoms/Widgets/EntityReferenceAutocomplete.js
lauriii: I think we should force keep it always off 👍

PR #423, opened 2018-08-27
Use Loadable Components

mattgrill:
## Issue

Our bundle size is quite large, especially with `react-rte`. It's perhaps time to revisit this issue now that we have a more stable base.

## Screenshot / UI changes

Use `loadable-components` to break apart the bundle and only load the parts we need. Currently the splitting happens at two levels, the routes and the widgets level. 

## Testing instructions

Make sure things don't break.


martinfrances107:
This is a good improvement. 👍 

It is something that has been on my mind for a few weeks now....it is good to see some movement in this direction.

Because it is such an obvious first response to our issues I am happy to see this go in without any performance tests.

Here we are not addressing "error handling" or prefetching of routes  [ see for example the appropriate sections in  https://www.npmjs.com/package/loadable-components ]

but as a first cut this is good.  Other refinements can come later after this  architectural decision to start using the loadable-components module.

dawehner:
@drpal do you mind having some actual numbers here? What is it before / how
much can we improve?

justafish:
I don't think it's the right time to re-visit this myself, but if we are looking at it then I think it's better follow React Router and Reach Router's recommended method
https://reacttraining.com/react-router/web/guides/code-splitting
https://reach.tech/router/large-scale

dawehner:
For me optimizations needs to be tested against the current baseline. We
also need to think about other future usecases potentially.

justafish:
see also https://reactjs.org/docs/code-splitting.html which says that code splitting is already setup in CRA, but also goes on to suggest using react-loadable (same as react-router and reach router)

mattgrill:
@dawehner - some numbers. These are all from the development server, so built sizes are probably a bit smaller due to production React, etc.

With everything in a single bundle, Chrome reports the bundle size as `897KB`. Splitting out routes at widgets, the bundle is `719KB`. Visiting `/node/add` loads another `11.7KB`.

@justafish I had looked at `react-loadable` before. It's the tool that is mentioned in both of those examples. You can do custom loading components with `loadable-components`, [smooth-code/loadable-components#custom-loading](https://github.com/smooth-code/loadable-components#custom-loading).

mattgrill:
Some numbers after running the build command, 

Before,
```
302.03 KB  build/static/js/main.5a859d94.js
```

After, 
```
157.93 KB  build/static/js/main.92bf6e50.js
113.45 KB  build/static/js/10.caf5f933.chunk.js
10.9 KB    build/static/js/0.2c4208f8.chunk.js
9.12 KB    build/static/js/6.31bc4bc4.chunk.js
8.75 KB    build/static/js/3.976c934d.chunk.js
7.74 KB    build/static/js/4.f323fb7b.chunk.js
7.58 KB    build/static/js/5.8d36db5a.chunk.js
7.28 KB    build/static/js/2.524ec878.chunk.js
6.39 KB    build/static/js/1.20271633.chunk.js
1.9 KB     build/static/js/7.67a28bbf.chunk.js
1.12 KB    build/static/js/9.8b6450d6.chunk.js
991 B      build/static/js/11.7896465f.chunk.js
983 B      build/static/js/8.a678df66.chunk.js
514 B      build/static/js/13.56643fb4.chunk.js
477 B      build/static/js/12.8346789e.chunk.js
```

mattgrill:
To provide some more context about chunk sizes. These are named after their `Widget` or `Route`.

```
158.07 KB  build/static/js/main.c0bacb34.js
113.46 KB  build/static/js/TextTextarea.3d867d47.chunk.js
10.9 KB    build/static/js/Permissions.3296125d.chunk.js
9.13 KB    build/static/js/EntityReferenceAutocomplete.528f070a.chunk.js
8.75 KB    build/static/js/FileUploadWidget.ef28a603.chunk.js
7.74 KB    build/static/js/NodeEditForm.ba302e1d.chunk.js
7.59 KB    build/static/js/NodeAddForm.70782cf7.chunk.js
7.28 KB    build/static/js/Dblog.21df8f6e.chunk.js
6.39 KB    build/static/js/AddContent.ae962b7a.chunk.js
1.9 KB     build/static/js/Roles.e5afdec9.chunk.js
1.13 KB    build/static/js/BooleanCheckbox.997c9893.chunk.js
995 B      build/static/js/OptionsSelect.aa3e209d.chunk.js
989 B      build/static/js/NumberTextfield.a8c46c6f.chunk.js
522 B      build/static/js/DatetimeTimestamp.a1ac75f5.chunk.js
483 B      build/static/js/StringTextfield.e6852123.chunk.js
```

Additionally, @dawehner provided this which is the JavaScript loaded for the current D8 article form.

image

dawehner:
I think most node forms will have this expensive component. Given that I think we could massively reduce the amount of splits by splitting up per route instead of per widget. The other widgets are so small, that I don't think the additional JS bundles are worth it.

justafish:
The first two items alone added together are `291.7kb`, which is pretty close to what we have now (`302.0kb`), also with the added downside that gzip won't be able to work it's magic as well.

Seems like it'd be good to revisit this when we've got more stuff in there

mattgrill:
@justafish I think the benefit here is that you no longer need to load `302.0kb` if you don't need to visit the pages that result in the full bundle size.

Issue #378, opened 2018-08-21
Create entity structure scaffolding at the NodeEdit or NodeAdd component level.


mattgrill:

dawehner:
I'm gonna give this a try :)

PR #425, opened 2018-08-27
Move the createEntity call to the node add form

dawehner:
Solves https://github.com/jsdrupal/drupal-admin-ui/issues/378

### DIFF
--- a/src/components/05_pages/NodeAddForm/index.js
+++ b/src/components/05_pages/NodeAddForm/index.js
@@ -5,8 +5,30 @@ import LoadingBar from 'react-redux-loading-bar';
 import NodeForm from '../NodeForm';
 import { contentAddChange, contentAdd } from '../../../actions/content';
 import PageTitle from '../../02_atoms/PageTitle/PageTitle';
+import { createEntity } from '../../../utils/api/schema';
+import { requestSchema } from '../../../actions/schema';
+import SchemaPropType from '../NodeForm/SchemaPropType';
 
 class NodeAddForm extends React.Component {
+  static propTypes = {
+    bundle: PropTypes.string.isRequired,
+    contentAdd: PropTypes.func.isRequired,
+    entityTypeId: PropTypes.string.isRequired,
+    schema: PropTypes.oneOfType([SchemaPropType, PropTypes.bool]),
+    requestSchema: PropTypes.func.isRequired,
+  };
+
+  static defaultProps = {
+    schema: false,
+  };
+
+  componentDidMount() {
+    this.props.requestSchema({
+      entityTypeId: this.props.entityTypeId,
+      bundle: this.props.bundle,
+    });
+  }
+
   onSave = entity => {
     const data = {
       ...entity,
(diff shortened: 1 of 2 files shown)


src/components/05_pages/NodeAddForm/index.js
```diff
@@ -17,30 +39,32 @@ class NodeAddForm extends React.Component {
...
-        <LoadingBar />
-        <NodeForm {...this.props} onSave={this.onSave} />
-      </Fragment>
+      this.props.schema && (
+        <Fragment>
+          <PageTitle>Create {this.props.bundle}</PageTitle>
+          <LoadingBar />
+          <NodeForm
+            {...this.props}
+            entity={createEntity(this.props.schema)}
```
mattgrill: Should we add the requests for the default authored-by user and the created-on time to this?

dawehner: I guess that would make sense, wouldn't it?

src/components/05_pages/NodeAddForm/index.js
```diff
@@ -38,19 +46,27 @@ class NodeAddForm extends React.Component {
   };
 
   render() {
-    return (
-      this.props.schema && (
+    let result = null;
+    if (this.props.schema) {
+      const entity = createEntity(this.props.schema);
```
dawehner: How about moving all this into a helper method to make it more readable?

src/components/05_pages/NodeAddForm/index.js
```diff
@@ -38,19 +46,27 @@ class NodeAddForm extends React.Component {
   };
 
   render() {
-    return (
-      this.props.schema && (
+    let result = null;
+    if (this.props.schema) {
```
dawehner: This needs to check for the existence of the current user

src/components/05_pages/NodeAddForm/index.js
mattgrill: 👍

src/components/05_pages/NodeAddForm/index.js
mattgrill: @dawehner Since this is `NodeAdd` would there ever be an existing user?

dawehner: Maybe that, but also because otherwise the setting of the default user doesn't work

src/components/05_pages/NodeForm/index.js
```diff
@@ -67,11 +62,12 @@ class NodeForm extends React.Component {
   };
 
   componentDidMount() {
-    this.props.requestUser(1);
-    this.props.requestSchema({
-      entityTypeId: this.props.entityTypeId,
-      bundle: this.props.bundle,
-    });
+    if (!this.props.schema) {
+      this.props.requestSchema({
```
lauriii: Should we at least add a todo to remove this? It would be nice if this would be consistently done on the parent component.

dawehner: Good question. I'm happy with either way. Having less data fetching inside NodeForm itself is certainly not a bad idea

lauriii: I think this direction makes even more sense in the long run when we move towards supporting other entity types 🙂

mattgrill: This will be somewhat difficult to achieve. We don't currently capture the bundle in the route, just the `NID`. You could compute the bundle by looking at the `type` of the entity, however that's sort of fragile.

lauriii: Agreed that this should happen in a follow-up! Let's open the issue and add a todo comment. 👍

Issue #421, opened 2018-08-27
Content confirmation dialog contains incorrect styles.


mattgrill:
Part of the confirmation dialog contains unstyled text. This looks like it's due to the removal of `<DialogContentText />`

![content](https://user-images.githubusercontent.com/1090713/44669400-a3bc7d00-a9d4-11e8-8984-966a95f550fa.png)

anjiv:
Hi @mattgrill ,
I would like to work on this issue. Can you please assign this to me ?

PR #426, opened 2018-08-28
issue-421: Confirmation dialog text styling fixed.

anjiv:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/421

## Screenshot / UI changes

screen shot 2018-08-29 at 12 36 29 pm
Changed the list style to bullets and removed '-' from list items in confirmation dialog.

## Testing instructions

1. Select the content on which action needs to be performed.
2. On Actions tab, select any action e.g : Delete content.
3. Click Apply button.
4. Confirmation dialog list contains bullets instead of '-'.

### DIFF
--- a/src/components/05_pages/Content/ConfirmationDialog.js
+++ b/src/components/05_pages/Content/ConfirmationDialog.js
@@ -11,10 +11,15 @@ import Slide from '@material-ui/core/Slide';
 import List from '@material-ui/core/List';
 import ListItem from '@material-ui/core/ListItem';
 import ListItemText from '@material-ui/core/ListItemText';
+import Typography from '@material-ui/core/Typography';
 
 const styles = {
   dialogActionName: css`
     text-transform: lowercase;
+  `,
+  listItemText: css`
+    list-style-type: disc;
+    display: list-item;
   `,
 };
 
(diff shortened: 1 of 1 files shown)


mattgrill:
@anjiv Does this also fix the issue with `this action cannot be undone` being unstyled? I don't see it in your screenshot.

anjiv:
@mattgrill Thanks for flagging this. Missed to do that change. I will fix that and update the PR along with the screenshot.

anjiv:
@mattgrill Can you please suggest the styling of "This action cannot be undone". Because I found that the message is not wrapped in any material ui class as we are using material design and that's why no styling is applied. And as per my suggestion we can use Roboto font which is used globally and can increase its font size and make the text color Red as it's an alert message. So, what are your suggestions on this ??

lauriii:
@anjiv I think it would be fine to just wrap it with the `Typography` component which will make the font-family and font-size consistent. Any other adjustments could be done in a follow-up issue.

anjiv:
@lauriii That sounds great !! Thank you. I will update the PR with the changes.

lauriii:
Thanks for making the changes!

PR #427, opened 2018-08-28
Moves Redux Devtools Extension into dev dependencies.

smurrayatwork:
## Changes:

This moves the Redux Devtools Extension module to only be a dev dependency.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -10,6 +10,7 @@ import {
 } from 'react-router-redux';
 import { Provider } from 'react-redux';
 import createSagaMiddleware from 'redux-saga';
+// eslint-disable-next-line import/no-extraneous-dependencies
 import { composeWithDevTools } from 'redux-devtools-extension/developmentOnly';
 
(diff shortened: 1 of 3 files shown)


src/App.js
```diff
@@ -10,6 +10,7 @@ import {
 } from 'react-router-redux';
 import { Provider } from 'react-redux';
 import createSagaMiddleware from 'redux-saga';
+// eslint-disable-next-line import/no-extraneous-dependencies
 import { composeWithDevTools } from 'redux-devtools-extension/developmentOnly';
```
mattgrill: Even though this was part of dependencies, the full version was probably not being included in the final bundle. Have you seen? [zalmoxisus/redux-devtools-extension#14-using-in-production](https://github.com/zalmoxisus/redux-devtools-extension#14-using-in-production)

mattgrill:
Closing this for now. Perhaps in the future we may want to have `logOnly` on for production.

Issue #424, opened 2018-08-27
Drupal Europe discussion plan


dawehner:
Here are a couple of questions we should maybe try to address during Drupal Europe

* How to collect feedback from people related to the demo?
* How will we integrate the app into Drupal? How to get feedback and signoff from the core maintainers?
* How can we improve the test coverage?
* Roadmap and the goals for the next 6 months / 12 months
  * Some ideas:
    * Media integration
    * Creating / managing taxonomy terms
    * Better comment management UI
    * Localization of the UI
    * Translation support
    * Workflow support
    * Revisions support
    * Stabilizing the content authoring part
* **Find and tag issues as "Good first issue" for the sprint day to get more people involved?**
* Enable login in via the admin UI

mattgrill:
- Field level constraints.

dawehner:
It is done and was great

PR #428, opened 2018-08-28
fixes #384

### LINKED ISSUE #384: Saving nodes without image on image field when cardinality is set to 1 causes an error
On Umami installation profile the article node type has image field with cardinality set to allow uploading single image. If a user doesn't upload an image to that field and tries to save a node, it causes the following error:
screen shot 2018-08-22 at 00 17 37

The field is not configured to be required so that shouldn't be the cause for the problem.

Here's a gif to help to reproduce the problem:
image_upload_empty_error

mattgrill:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/384

_fwiw_, we really need to talk about a constraints system at Drupal Europe.

## Testing instructions

- make an article, don't upload an image, actually save the node.

### DIFF
--- a/src/components/05_pages/NodeAddForm/index.js
+++ b/src/components/05_pages/NodeAddForm/index.js
@@ -8,12 +8,34 @@ import PageTitle from '../../02_atoms/PageTitle/PageTitle';
 
 class NodeAddForm extends React.Component {
   onSave = entity => {
-    const data = {
-      ...entity,
-      type: `${this.props.entityTypeId}--${this.props.bundle}`,
-    };
-    this.props.contentAdd(data, this.props.bundle);
+    this.props.contentAdd(
+      this.cleanupRelationships({
+        ...entity,
+        type: `${this.props.entityTypeId}--${this.props.bundle}`,
+      }),
+      this.props.bundle,
+    );
   };
+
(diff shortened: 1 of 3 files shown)


dawehner:
> fwiw, we really need to talk about a constraints system at Drupal Europe.

Add it to our list on https://github.com/jsdrupal/drupal-admin-ui/issues/424

src/components/05_pages/NodeAddForm/index.js
```diff
@@ -8,13 +8,35 @@ import PageTitle from '../../02_atoms/PageTitle/PageTitle';
...
+    this.props.contentAdd(
+      this.cleanupRelationships({
+        ...entity,
+        type: `${this.props.entityTypeId}--${this.props.bundle}`,
+      }),
+      this.props.bundle,
+    );
   };
 
+  cleanupRelationships = ({ relationships, ...rest }) => ({
```
dawehner: Are you sure this is not a problem on node edit too?

mattgrill:
@dawehner

src/components/05_pages/NodeAddForm/index.js
mattgrill: It is.

PR #429, opened 2018-08-29
Enable restoring lost edited content

dawehner:
Follow up for https://github.com/jsdrupal/drupal-admin-ui/pull/357

resolves https://github.com/jsdrupal/drupal-admin-ui/issues/244

### DIFF
--- a/src/components/05_pages/NodeEditForm/index.js
+++ b/src/components/05_pages/NodeEditForm/index.js
@@ -6,6 +6,7 @@ import { css } from 'emotion';
 import NodeForm from '../NodeForm';
 import { contentSave, requestSingleContent } from '../../../actions/content';
 import PageTitle from '../../02_atoms/PageTitle/PageTitle';
+import { cleanupRelationships } from '../../../utils/api/content';
 
 let styles;
 
@@ -15,11 +16,12 @@ class NodeEditForm extends React.Component {
   }
 
   onSave = bundle => entity => {
-    const data = {
-      ...entity,
-      type: `${this.props.entityTypeId}--${bundle}`,
-    };
-    this.props.contentSave(data);
+    this.props.contentSave(
+      cleanupRelationships({
+        ...entity,
+        type: `${this.props.entityTypeId}--${bundle}`,
+      }),
+    );
   };
 
   render() {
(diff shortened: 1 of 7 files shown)


src/components/05_pages/NodeEditForm/index.js
```diff
@@ -77,12 +81,34 @@ styles = {
   `,
 };
 
+const extractRestorableEntity = (state, entity) => {
+  const restorableEntity = state.content.restorableContentEditById[entity.id];
+  if (
+    // @todo maybe that's too defensive?
```
lauriii: I think this looks fine to me. I'm wondering if we should have some kind of helper function for this type of conditions.

src/components/05_pages/NodeEditForm/index.js
```diff
@@ -77,12 +81,34 @@ styles = {
...
+    restorableEntity &&
+    restorableEntity.data &&
+    restorableEntity.data.attributes &&
+    entity &&
+    entity.attributes &&
+    // Just offer to restoring when the loaded entity wasn't changed in the meantime.
```
lauriii: Maybe `Restoring content is only offered when the loaded entity wasn't changed in the meantime.` would read a bit better.

src/reducers/content.js
```diff
@@ -27,15 +29,23 @@ export default (state = initialState, action) => {
...
-          ...state.contentAddByBundle,
+        restorableContentAddByBundle: {
+          ...state.restorableContentAddByBundle,
           [action.payload.bundle]: action.payload.entity,
         },
       };
+    case CONTENT_EDIT_CHANGE:
+      return {
+        ...state,
+        restorableContentEditById: {
```
dawehner: I'm wondering whether storing every editing content over time is the right approach. Maybe we should limit the amount of stored entires to let's say the last three?

lauriii: Let's open an issue and figure out the correct logic for that. I think for now this is perfectly fine.

dawehner: 👍 Let's be pragmatic indeed

dawehner: Here is one: https://github.com/jsdrupal/drupal-admin-ui/issues/432

src/components/05_pages/NodeEditForm/index.js
dawehner: That's indeed true. Thank you!

src/components/05_pages/NodeEditForm/index.js
dawehner: Yeah I guess we could have one of those. The same probably for setters.

Issue #430, opened 2018-08-29
Multiple Fields validations issue and incorrect mapping


vishalchandna1:
<img width="1051" alt="screen shot 2018-08-29 at 1 32 26 am" src="https://user-images.githubusercontent.com/13252288/44770001-2899c900-ab84-11e8-8010-d2e85f33f1cd.png">

1. A validation needs to be added when checking out the multiple fields values. The whole object i.e `values` props needs to be validated (currently we are validating only the length as `values.length`)
 
2. The mapping must be done of the newly created array i.e usedValues, as we are copying the data from the `values` props and adding it to the `userValues` with a validation

vishalchandna1:
Sure. Thanks

PR #431, opened 2018-08-29
MultipleFields usedValues mapping and values.length validation correc…

vishalchandna1:
…ted.

## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/430

`values.length` validation changed to the whole object validation

`values.map` changed to `usedValues.map` to render an empty ingredient field on **add recipe** page. It also resolves the error for the **node edit page**

### DIFF
--- a/src/components/02_atoms/MultipleFields/MultipleFields.js
+++ b/src/components/02_atoms/MultipleFields/MultipleFields.js
@@ -171,7 +171,9 @@ class MultipleFields extends Component {
       state: { newItemAdded },
       props: { label, value: values, component, onChange },
     } = this;
-    const usedValues = (values && values.length) || [this.createEmptyItem()];
+    const usedValues = (values && values.length && values) || [
+      this.createEmptyItem(),
+    ];
     return (
       <FormControl margin="normal" fullWidth>
         <FormLabel component="legend">{label}</FormLabel>
--- a/src/components/02_atoms/MultipleFields/MultipleFields.js
+++ b/src/components/02_atoms/MultipleFields/MultipleFields.js
@@ -171,6 +171,9 @@ class MultipleFields extends Component {
       state: { newItemAdded },
       props: { label, value: values, component, onChange },
     } = this;
+
+    // values && values.length is to validate the object is not null and not an empty array, respectively
+    // the last `&& values` is to make sure after validation, it always return `values`  instead of `values.length`
     const usedValues = (values && values.length && values) || [
       this.createEmptyItem(),
     ];
--- a/src/components/02_atoms/MultipleFields/MultipleFields.js
+++ b/src/components/02_atoms/MultipleFields/MultipleFields.js
@@ -171,7 +171,7 @@ class MultipleFields extends Component {
       state: { newItemAdded },
       props: { label, value: values, component, onChange },
     } = this;
-    const usedValues = values || [this.createEmptyItem()];
+    const usedValues = (values && values.length) || [this.createEmptyItem()];
     return (
       <FormControl margin="normal" fullWidth>
         <FormLabel component="legend">{label}</FormLabel>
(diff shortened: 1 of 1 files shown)


src/components/02_atoms/MultipleFields/MultipleFields.js
```diff
@@ -171,13 +171,13 @@ class MultipleFields extends Component {
...
     } = this;
-    const usedValues = values.length ? values : [this.createEmptyItem()];
+    const usedValues = values || [this.createEmptyItem()];
     return (
       <FormControl margin="normal" fullWidth>
         <FormLabel component="legend">{label}</FormLabel>
         <List>
           {usedValues &&
-            values.map((value, index) => (
+            usedValues.map((value, index) => (
```
dawehner: oh yeah that's indeed needed!

src/components/02_atoms/MultipleFields/MultipleFields.js
```diff
@@ -171,13 +171,13 @@ class MultipleFields extends Component {
       state: { newItemAdded },
       props: { label, value: values, component, onChange },
     } = this;
-    const usedValues = values.length ? values : [this.createEmptyItem()];
+    const usedValues = values || [this.createEmptyItem()];
```
dawehner: Are you sure about this particular change? An empty array is trueish: ```> const a = [];
undefined
> !!a;
true```

vishalchandna1: Yes, I have tested that. We are either receiving the `values` as `null` or an array with length > 0
Do you want me to add the array validation?

src/components/02_atoms/MultipleFields/MultipleFields.js
```diff
@@ -171,13 +171,13 @@ class MultipleFields extends Component {
       state: { newItemAdded },
       props: { label, value: values, component, onChange },
     } = this;
-    const usedValues = values.length ? values : [this.createEmptyItem()];
+    const usedValues = (values && values.length) || [this.createEmptyItem()];
```
vishalchandna1: @dawehner 
I have added the array validation also. It would make it robust. Thanks for your feedback :)

dawehner: 👍 for more robustness!

src/components/02_atoms/MultipleFields/MultipleFields.js
```diff
@@ -171,13 +171,15 @@ class MultipleFields extends Component {
       state: { newItemAdded },
       props: { label, value: values, component, onChange },
     } = this;
-    const usedValues = values.length ? values : [this.createEmptyItem()];
+    const usedValues = (values && values.length && values) || [
```
vishalchandna1: `values && values.length && values` needed to be added in order to return `values` instead of `values.length` 

`values && values.length` returns `values.length`

dawehner: why would you need both values && values twice

dawehner: oh I see what you did, mh, at least I was able to understand 1. on the first read

vishalchandna1: first `values` to validate if the object is null or not, 
then `values.length` to check if the array is empty or not
lastly 'values' is to return the value as `values` instead of the `values.length`

if it would have been `values && values.length`, it would return the value as `values.length` instead of the actual object which is `values`

dawehner: Well, I totally understand how it works, no worries. The problem is someone new to the project trying to understand the code. It feels like 1. which is a bit more explicit would be the nicer alternative overall, but that's maybe just me.

mattgrill: This is a clever solution. Nice job @vishalchandna1. However. I think that the first solution is probably a bit more explicit and easier to understand. Perhaps we could add just a bit of documentation about what is happening, and then sticking with the current solution seems good.

dawehner: Good idea!

vishalchandna1: Got it, I will add it right away.

vishalchandna1:
@dawehner @mattgrill  Added the docs for it.

Issue #433, opened 2018-08-29
Reduce complexity:  One Dialog element per table.


martinfrances107:
Regarding the OpsModalButton, I think I casually introduced a pattern which is responsible for slow page loading which I now want to refactor.

The current problem is that we have one hidden dialog associated with every table row, or for every button

Each DIALOG is a structure with 12 base html element ( BUTTON, DIV, P,, SPAN etc )  

This has scaling problems, currently the list content page has 12 rows so that is 144 elements.

I can think of projects where I want 25 rows per table. ( 300 elements ) 

I have created a PR with a new OpsTableElement which contains centralized DIALOG element.
all table sub elements are wrapped in a react "context"  so all child  OpsModalButton's can call a openDIalog method held within that  context which populates the dialog and then unhides it.

I have the skeleton of the code worked out ... it has a few bugs ... 
on the storybook page only click the OpsModalButton the Dialog opens, it can be closed... but not opened a second time., but this is work in progress so I am just showing the idea...if anyone has objections then it can be restructured or shot down early.

dawehner:
I would measure first here :) measure what it takes to render right now,
make improvements and look at it again. I think conceptually what you write
make sense and rendering less Dialogs would help a lot.

martinfrances107:
With the next iteration I will provide timing evidence .. good or bad.

But by way of providing supporting evidence -- I want to advance the line that we should do this even if the timing is not significantly improved - because of scaling concerns.

Chrome's lighthouse auditing tools: in the performance section. There are comments about an overly complicated DOM structure 

The total number of DOM elements is about 80% of the failure threshold ( which is a pass ) 

BUT we are getting warnings saying 

The number of descendant children is too large 

html  ...  table->tbody->trow->tcell -> before our dialog and its 12 elements.

I think other developers in contrib are going to want to create far more complex tables that the one we are basing our design around. Certainly many are going to want to set the pagination upto to a table with 25 or 50 rows ... and which point as I tried to point out in the main issue .. the table structure does not scale. The  20%  headroom before the DOM element count fails - will be easily consumed.

Regarding performance I want to highlight a second issue. Well for me the original issue.

The animation performance of the side drawer opening is a horrendous 2-4 fps.  ( on a low end laptop ) so mobile should be worse.

I have pofiled the animation  .. and the cause is a per frame resizing of the table as the drawer opens.
[...]

PR #434, opened 2018-08-29
WIP Define a single Dialog inside OpsTable

martinfrances107:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/433

This is a refactoring of the OpsModalButton.

Gone is the pattern of one DIalog per table row, now we have a single dialog defined above the table -  which the button can populate with text and display.

( It reduces the page's DOM element count scaling problem, and is a react style lazy evaluation pattern )  

## Testing instructions

Goto the storybook definition of OpsTable...

Please note there is a bug as described in #433 The button opens the dialog as expected 
the dialog can be edited or cancelled BUT will not open a second time .. this is WIP and I am working on it.


martinfrances107:
I have added a series of minor code corrections. I will hold off on the performance discussion until the bug is fixed.

martinfrances107:
I have now fixed the bug.... 

as per @dawehner suggestion I will do a performance report .. but it too late tonight.. in the morning.

martinfrances107:
So here is the performance discussion:

The summary is that we are at an early stage of development, and there are a few buggy aspects of the code that prevent the generation of meaningful numbers.. at least ones I trust.

So my template for Lighthouse auditing.

a) switch to incognito mode.
b) run test with cold cache and warm cache.
c) rerun tests a few times to gauge consistency

but when I try and test on the production /vfancy version of the site... I failed.

1) I could not use incognito mode -- as we are in a rapid development state and we have a CORS issue, in that we need a tab with localhost:8888 running in parallel otherwise request for menu data fails.

2) I  could not use /vfancy for a second reason localhost:8888/vfancy redirects to localhost:8888 when the page is refreshed... which unfortunately means lighthouse cannot be set to run on the vfancy page it gets confused -- refreshes and reports on localhost:/8888.

When I try running lighthouse in a **highly unrecommended** mode using localhost:3000. - The variability of runs on the same page is too high to say anything with confidence about a final production build.

But which that said the 

**First contentful paint** improves from 7,720 ish to 7,100 ms with error bars +/- 100ms on each value.

**First meaningful paint** shows maybe a similar improvement of maybe 500ms
 
BUT I really don't trust my findings ... I think running on vfancy is critical and I should open another issue to get testing on production like builds

martinfrances107:
As described in the issue ... this PR is wrong headed the complexity I wanted to take out 
does not improve the performance, nor will it improve the scalability as the high compute cost is only born by dialogs rendered on the page .. the two ways of working have the same DOM element count.

Issue #437, opened 2018-08-30
TypeError on widgets.


anjiv:
<img width="992" alt="screen shot 2018-08-30 at 12 54 14 pm" src="https://user-images.githubusercontent.com/25002064/44836174-d1632980-ac53-11e8-9773-9fc16c11089f.png">

anjiv:
@dawehner I have updated the PR and added the documentation for object shape. Thanks for the heads up regarding the documentation :)

dawehner:
Thank you @anjiv

PR #438, opened 2018-08-30
issue-437: TypeError resolved on widgets.

anjiv:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/437

## Testing instructions

1. Go to the URL : /node/add/recipe.
2. Check the console. The TypeError related to widgets.option_select on NodeForm is not reproducible now.

### DIFF
--- a/src/components/05_pages/NodeForm/index.js
+++ b/src/components/05_pages/NodeForm/index.js
@@ -29,6 +29,11 @@ class NodeForm extends React.Component {
       PropTypes.oneOfType([
         PropTypes.func,
         PropTypes.instanceOf(React.Component),
+        PropTypes.shape({
+          options_select: PropTypes.shape({
+            component: PropTypes.instanceOf(React.Component),
+          }),
+        }),
       ]).isRequired,
     ).isRequired,
     onSave: PropTypes.func.isRequired,


dawehner:
Thank you for the pull request!

How about documenting that it can be a shape like ```shape({ . component: ..., multiple: boolean})```?

src/components/05_pages/NodeForm/index.js
```diff
@@ -29,6 +29,9 @@ class NodeForm extends React.Component {
       PropTypes.oneOfType([
         PropTypes.func,
         PropTypes.instanceOf(React.Component),
+
+        // Our object shape will be like PropTypes.shape({field_name: PropTypes.shape({component: PropTypes.instanceOf(React.Component), multiple: PropTypes.bool})}).
```
dawehner: Oh I'm sorry, I could have been more clear. You could have used PropTypes.shape in the actual proptype :) Does that make sense?

anjiv:
@dawehner I have updated the PR and have done it for option_select only for now and we need to do this for other widgets also. So, should I add that in this PR itself or we will cover in other follow up issue ?? 
We need to add for specific widgets because I tried making it generic with this approach : 
```
PropTypes.objectOf(PropTypes.shape({
   component: PropTypes.instanceOf(React.Component),
}))
```
But it didn't worked out. So, we need to add for specific widgets. Please correct me if I am getting wrong anywhere.

mattgrill:
@anjiv - I think it's fine to keep this just for `OptionsSelect` and we can open up a followup issue to fix the PropTypes on the other components.

anjiv:
@mattgrill Okay. Thank you. Then, we are good to go on this issue as I have resolved it for `OptionsSelect`.

src/components/05_pages/NodeForm/index.js
```diff
@@ -29,6 +29,11 @@ class NodeForm extends React.Component {
       PropTypes.oneOfType([
         PropTypes.func,
         PropTypes.instanceOf(React.Component),
+        PropTypes.shape({
```
dawehner: I'm a bit confused. I would have expected it would be a ```PropTypes.oneOfType([PropTypes.func, PropTypes.instanceOf(React.Component), PropTypes.shape({component: PropTypes.instanceOf(React.component), multiple: PropTypes.booln})])```

anjiv: @dawehner I also thought of this but `component` is available inside `options_select`. That's why implemented it in this way.

dawehner: Looking at https://github.com/jsdrupal/drupal-admin-ui/blob/master/src/components/05_pages/NodeForm/Widgets.js#L11 it looks like all of them have a component now :)

anjiv: Yeah you are right here, that's why we need to implement this part of code which I implemented for `options_select` for other widgets too. And @mattgrill suggested that we can cover that in follow up ticket. So, for now I did for `options_select` only.

dawehner:
A recent pull request removed the need to have a prop type here. Sorry for the inconveniences.

Issue #440, opened 2018-08-30
Cut - off unnecessary API hits for title filtering on Content page


vishalchandna1:
For the title filtering on content, a request is fired on every character change which is unnecessary and creating a stuttery user experience while typing. e.g If I need to search a title `baking` then it would fire 6 API requests ['b', 'ba', 'bak', 'baki', 'bakin', ''baking'] and it doesn't make sense to fire off all the requests to reach our end goal.
![baking](https://user-images.githubusercontent.com/13252288/44838655-4c2f4300-ac5a-11e8-8111-51f88dcea908.gif)

we need to cut-off all other unnecessary requests in order to efficiently utilize the API and provide a seamless user experience.

vishalchandna1:
**Without optimizations**
https://drive.google.com/file/d/1E7QPmaLCdYLLz1UvCpSuIM8SDiukJPLD/view
**With Optimizations**
https://drive.google.com/file/d/1JG3L5-lCX50W2qA2x32rwv73u6Du4eSF/view

If you also see, the loading is also consistent with delay and it doesn't go back and forth when a user is typing.

dawehner:
This is much nicer! Thank you @vishalchandna1 
Do you think half a second is okay in terms of waiting time? I would have expected a lower number to be honest.

vishalchandna1:
Yep, we can do some experiments here to find the optimum value.

vishalchandna1:
@dawehner  I have experimented with it a lot and we can go for 400ms. Going below that may cause a few more API calls than usual, which depends on the typing speed of the user. 
It works seamlessly, Even with 500ms the delay is noticeable at all. I have updated the PR with 400ms too :)

PR #441, opened 2018-08-30
Add delay and cut-off extra api requests.

vishalchandna1:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/440

untitled

### DIFF
--- a/src/components/05_pages/Content/Content.js
+++ b/src/components/05_pages/Content/Content.js
@@ -175,7 +175,17 @@ class Content extends Component {
                         className={styles.textField}
                         onChange={e => {
                           this.setState({ title: e.target.value }, () => {
-                            this.props.requestContent(this.state);
+                            const { title } = this.state;
+
+                            // Clearing up the previous request if user is entering some more letters.
+                            if (this.requestContentWithDelay) {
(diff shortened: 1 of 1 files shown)


vishalchandna1:
Hi @mattgrill 

We have our current implementation with `takeLatest()` itself and it doesn't really stop the API hits, however it keeps track of the very latest action.

Also, I have tried doing it by `throttle` effect offered by `react-saga`, it works but not perfectly. I have to provide it with a value of **2000ms** to make it work but that is also not up to the mark. 

https://github.com/redux-saga/redux-saga/tree/master/docs/recipes

With throttle integrated, The API is always fired **when the first letter is entered/removed** and after that, it works fine. I believe that this is a bug from in the `throttle` effect itself as it doesn't keep track of the first action which is further taken to the reducer.

What are your thoughts on the current code?

vishalchandna1:
Hi @lauriii 
I think we should keep our dependencies as minimal as possible and with this package also we would have to implement the cut-off API stuff.

dawehner:
@mattgrill and @dawehner had a look at that and came up with an alternative solution: https://github.com/jsdrupal/drupal-admin-ui/pull/455

PR #445, opened 2018-08-30
Entity type of referenced entities is not persisted on state update

lauriii:


lauriii:
> This happens when you select multiple referenced entities?

Yes

dawehner:
This looks good for me!

PR #450, opened 2018-08-31
Cleanup how route params are handled.

mattgrill:
## Issue

Routes/Components that need props from React Router are handled a bit different than those that do not. There is no reason for this. This moves the route matching params logic to either `NodeEdit` or `NodeAdd` instead of the actual router.

### DIFF
--- a/src/components/05_pages/NodeForm/index.js
+++ b/src/components/05_pages/NodeForm/index.js
@@ -20,17 +20,13 @@ import { setErrorMessage } from '../../../actions/application';
 
 import { createUISchema, sortUISchemaFields } from '../../../utils/api/schema';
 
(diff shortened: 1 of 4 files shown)


src/components/05_pages/NodeForm/index.js
```diff
@@ -20,17 +20,13 @@ import { setErrorMessage } from '../../../actions/application';
 
 import { createUISchema, sortUISchemaFields } from '../../../utils/api/schema';
 
+import widgets from './Widgets';
```
dawehner: Seems fine, this information will always live in JS anderen Not be dynamic

dawehner:
We can always iterate later when we need this components without the router

PR #452, opened 2018-09-04
Use saga/effects all instead of Promise.all()

mattgrill:
## Issue

We're currently using `Promise.all()` instead of `all([])` from `redux-saga`. https://redux-saga.js.org/docs/api/#alleffects---parallel-effects

## Testing instructions

- Ensure schema is still fetched correctly.


martinfrances107:
Yep, this looks good 

The concept is sound, 

I have pull and tested ... yes this should go in.

Issue #487, opened 2018-09-17
Nightwatch Tests Failures


mattgrill:
Sometimes `/node/9/edit` is not `Super easy vegetarian pasta bake`

mattgrill:
Fixed!

PR #488, opened 2018-09-17
Fix flaky nightwatch tests.

mattgrill:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/487


dawehner:
Sadly it still fails :|

mattgrill:
@dawehner Now it passes.

dawehner:
It feels like it would be nicer to use an ID for the data-nightwatch attribute, but ¯\_(ツ)_/¯

PR #500, opened 2018-10-02
Update to Create react app 2.0

dawehner:
* Followed the official update instructions
* Manually tested adding an article and updating a recipe


package.json
```diff
@@ -53,15 +53,15 @@
     "@storybook/addons": "^3.4.8",
     "@storybook/react": "^3.4.8",
     "babel-core": "^6.26.3",
```
justafish: Should this be updated too?

lauriii: I tried to upgrade babel to the latest version but it seems like there's not version of storybook that would allow upgrading babel.

@justafish Are you fine if we push this to a follow-up PR?

martinfrances107: Just looking at the output of yarn outdated, and want to talk about the scope of the follow-up.

There does seem to be a lot of red. I appreciate there is a balance between stability and being current. My way of thinking ... is that it is better to bite the bullet often and be responsive to issues that come up, rather than get stuck in intractable log jams.

For me material-ui is the stand out thing that is getting well behind the curve.

outdated

dawehner:
Finally rerolled

package.json
```diff
@@ -53,16 +53,16 @@
...
     "babel-preset-env": "^1.7.0",
     "babel-preset-react-app": "^3.1.2",
     "babel-register": "^6.26.0",
     "babel-runtime": "^6.26.0",
     "chromedriver": "^2.40.0",
     "enzyme": "^3.3.0",
     "enzyme-adapter-react-16": "^1.1.1",
-    "eslint": "^4.9.0",
-    "eslint-config-airbnb": "^16.1.0",
+    "eslint": "5.6.0",
```
dawehner: Note: CRA requires to update eslint, which requires us to update eslint-config-airbnb which enables additional eslint rules. Some of them I disabled above

package.json
```diff
@@ -92,5 +92,11 @@
     "test:unit": "CI=true JEST_JUNIT_OUTPUT=reports/jest-junit.xml react-scripts test --testResultsProcessor ./node_modules/jest-junit --env=jsdom",
     "test:lint": "eslint ./src --max-warnings=0 --format junit -o reports/eslint-junit.xml",
     "eject": "react-scripts eject"
-  }
+  },
+  "browserslist": [
```
dawehner: This is the default suggestion, seems sensible for now, but we can always update.

Issue #502, opened 2018-10-10
Upgrade jest-junit 


martinfrances107:
Followup to #501 

just executing the following line a fixing up the test failures.

yarn upgrade   jest-junit --latest

Jest-junit is two major version numbers behind 3.7 -> 5.2

when I run 

yarn test

I see lots of deprecation warnings from #501 regarding material-ui's ( all typography related ) 

I am not sure if this is because updating junit has made it more verbose or what...

```
    console.error node_modules/@material-ui/core/node_modules/warning/warning.js:36
      Warning: Material-UI: You are using the deprecated typography variant subheading that will be removed in the next major release. 
      Please read the migration guide under https://material-ui.com/style/typography#migration-to-typography-v2
    console.error node_modules/@material-ui/core/node_modules/warning/warning.js:36
      Warning: Material-UI: You are using the typography variant body1 which will be restyled in the next major release. 
      Please read the migration guide under https://material-ui.com/style/typography#migration-to-typography-v2
    console.error node_modules/@material-ui/core/node_modules/warning/warning.js:36
      Warning: Material-UI: You are using the deprecated typography variant subheading that will be removed in the next major release. 
      Please read the migration guide under https://material-ui.com/style/typography#migration-to-typography-v2
    console.error node_modules/@material-ui/core/node_modules/warning/warning.js:36
      Warning: Material-UI: You are using the typography variant body1 which will be restyled in the next major release. 
      Please read the migration guide under https://material-ui.com/style/typography#migration-to-typography-v2

```

In any event  there are some more test failures to fix as well... I will sort this out slowly over the next week.

martinfrances107:
OK it seems the deprecation warning will have to be dealt with by upgrading other components 

and the test failures I can recreate on a clean checkout  so I am just going to post the changes due to one command

yarn upgrade jest-junit --latest

dawehner:
We no longer need these manual update issues, yeah!

PR #503, opened 2018-10-10
#502 upgrade jest-junit to 5.2

martinfrances107:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/502

jest-junit is two major version numbers behind the lastest version ... 

time for an upgrade... This patch marks a transition from 3.7 to 5.2

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "eslint-plugin-jsx-a11y": "^6.1.0",
     "eslint-plugin-prettier": "^2.6.0",
     "eslint-plugin-react": "^7.4.0",
-    "jest-junit": "^3.6.0",
+    "jest-junit": "^5.2.0",
     "mkdirp": "^0.5.1",
     "nightwatch": "^0.9.21",
     "prettier": "^1.11.1",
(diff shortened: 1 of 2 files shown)


dawehner:
Isn't this coming from create-react-app automatically?

martinfrances107:
So facebook does have a recommended way of solving the outdated issue.

When I find the time I will start following this way of working.

https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#updating-to-new-releases

> To update an existing project to a new version of react-scripts, open the changelog, find the version you’re currently on (check package.json in this folder if you’re not sure), and apply the migration instructions for the newer versions.
> 
> In most cases bumping the react-scripts version in package.json and running npm install (or yarn install) in this folder should be enough, but it’s good to consult the changelog for potential breaking changes.
> 
> We commit to keeping the breaking changes minimal so you can upgrade react-scripts painlessly.

dawehner:
We no longer need these tiny updates. greekeeper will  do them for us. Thank you for your contribution

Issue #497, opened 2018-09-24
Epic: Component library


dawehner:
One of the goals for the admin UI inititaive is to provide a reusable component library for react so people can provide some custom apps with the same look and feel

## Designs

TODO

## Implementatioin

TODO
 
## Components

- [ ] Component A issue
- [ ] Component Bb issue

PR #504, opened 2018-10-12
Monorepo structure for allowing usage of components outside of this project

lauriii:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/497

## Description
This modifies the directory structure so that the project is set up as monorepo. I used yarn workspaces but we could consider using [lerna](https://github.com/lerna/lerna) if we find their tooling useful. The packages are under `packages/` folder. The new `drupal-ui` package will contain all of the reusable components.

To make sure that at least the build works, I created an entry point which contains all of the components inside the component library. We might want to modify this so that components can be loaded one by one.

## Remaining tasks
- Update documented installation steps to match with the new structure
- Update installation steps used in the CI to match with the documented steps
- Allow loading of individual components.

## Example of installation steps for developers wanting to use the component library
*Note: these steps won't work before the package is published in the npm registry*

Add dependency for the package 🚀
`yarn add drupal-ui`

Use components 👯‍♀️🎉
```
import { PageTitle } from 'drupal-ui';

const app = () => (
  <div>
    <PageTitle>Title</PageTitle>
  </div>
)
```

### DIFF
--- a/src/components/05_pages/TaxonomyTermsOverview/TaxonomyTermsOverview.js
+++ b/src/components/05_pages/TaxonomyTermsOverview/TaxonomyTermsOverview.js
@@ -18,17 +18,18 @@ import Button from '@material-ui/core/Button';
 import AddIcon from '@material-ui/icons/Add';
 
 import FormControl from '@material-ui/core/FormControl';
+import InputLabel from '@material-ui/core/InputLabel';
 import Select from '@material-ui/core/Select';
 import MenuItem from '@material-ui/core/MenuItem/MenuItem';
 
 import PageTitle from '../../02_atoms/PageTitle';
 
 const styles = {
-  addButton: css`
-    margin: 0.5rem;
-    position: fixed;
-    right: 0;
-    bottom: 0;
+  action: css`
+    margin-left: 0rem;
+    margin-bottom: 0.5rem;
+    min-width: 5rem;
+    max-width: 15rem;
   `,
 };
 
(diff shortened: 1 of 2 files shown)


packages/drupal-admin-ui/src/components/05_pages/AddContent/AddContent.js
```diff
@@ -8,7 +8,7 @@ import Paper from '@material-ui/core/Paper';
 import List from '@material-ui/core/List';
 import ListItem from '@material-ui/core/ListItem';
 import ListItemText from '@material-ui/core/ListItemText';
-import PageTitle from '../../02_atoms/PageTitle';
+import { PageTitle } from 'drupal-ui';
```
dawehner: I love how this makes things actually more readable!

packages/drupal-ui/build/InlineMessage/InlineMessage.js
```diff
@@ -0,0 +1,46 @@
+var _modifierClassMap;
```
dawehner: We  should not commit the build folder

packages/drupal-admin-ui/src/components/02_atoms/Widgets/BooleanCheckbox.js
```diff
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import CheckBox from '@material-ui/core/Checkbox';
 import FormControlLabel from '@material-ui/core/FormControlLabel';
 import { css } from 'emotion';
-import WidgetPropTypes from '../../../drupal-admin-ui/src/components/05_pages/NodeForm/WidgetPropTypes';
```
dawehner: I'm wondering whether it's a valid usecase to try to reuse these widgets?

lauriii: It seems like these widgets are more about functionality than the visual looks than the rest of the components so we probably should look at them separately at some point.

lauriii:
It seems like the `lint_unit` test is failing because of there's something wrong with CircleCI itself. I run the task using CircleCI locally and the task is passing. My local installation didn't support caches, so I would guess this problem to be most likely caused by that.

justafish:
Moved to #515

PR #509, opened 2018-10-25
Merge the drupal-admin-ui-demo project into this project.

dawehner:
- Merge `drupal-admin-ui-demo` project into this project.
- Adjust CircleCI build steps to accommodate this change.


lauriii:
Does this change the installation steps users has to take? Should we worry at all of the pre-existing installations or should we just tell them to re-install from scratch?

dawehner:
> Does this change the installation steps users has to take? Should we worry at all of the pre-existing installations or should we just tell them to re-install from scratch?

* It will change existing demo users
* It won't change how people work on the ```drupal-admin-ui```
* It just changes the way how people work on the demo piece.

Issue #519, opened 2018-11-02
react-axe: Bump version


martinfrances107:
Currently we are on 3.0.1 ... 3.0.2 is available 

I will get to this shortly.

PR #524, opened 2018-11-04
react-axe: Bump version #519

martinfrances107:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/519

Just a minor version number update.

### DIFF
--- a/packages/admin-ui/package.json
+++ b/packages/admin-ui/package.json
@@ -68,7 +68,7 @@
     "mkdirp": "^0.5.1",
     "nightwatch": "^0.9.21",
     "prettier": "^1.11.1",
-    "react-axe": "^3.0.1",
+    "react-axe": "^3.0.2",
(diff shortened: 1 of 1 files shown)


packages/admin-ui/package.json
```diff
@@ -68,7 +68,7 @@
     "mkdirp": "^0.5.1",
     "nightwatch": "^0.9.21",
     "prettier": "^1.11.1",
-    "react-axe": "^3.0.1",
+    "react-axe": "^3.0.2",
```
lauriii: Is there a specific reason we need to depend on this specific patch release of `react-axe`? Otherwise, wouldn't it be enough to run `yarn upgrade`?

martinfrances107: Thank you for looking at this, and asking questions ...

but I think #528  is the way to keep things in sync

Issue #64, opened 2018-03-31
Consider using Typescript


pixelmord:
I do not want to open up a huge discussion on pros and cons, but I thought to bring this up early would make sense.
I personally am doing my first project that will go into production with Typescript and saw a definite improvement in code quality and safety. It was hard in the beginning especially with some more complex definitions in React for interfaces of some HOCs of e.g. react-apollo, but in the end, it really pays off especially when an application gets more and more complex.
Since this here will no doubt be a complex beast, I would recommend using Typescript over just defining PropTypes...

I would gladly help by refactoring the current code with https://www.npmjs.com/package/react-scripts-ts if we can agree upon using TS

justafish:
Thanks for bringing this up! Whilst I think TypeScript (and Flow) are great. The thing that makes me hesitate the most is this:
https://insights.stackoverflow.com/survey/2018#technology
JavaScript is at 70% and TypeScript at 17%
I think we're already attempting to make a big leap for the Drupal community in technology terms, this might be a bit too much (for now). One of the biggest goals for me is making this comfortable and accessible to "regular" JavaScript developers.

It's nice to see it's one of the most loved languages though! https://insights.stackoverflow.com/survey/2018#technology-most-loved-dreaded-and-wanted-loved

dawehner:
We had a similar discussion for a new project we had a work. One thing we realized tools like typescript or flow exist as people experienced pain along the path and came up with a solution for it. 

This actually really similar to how redux for example came in place. People realized that local state managment can be hard, so they came up with the [flux architecture](http://fluxxor.com/what-is-flux.html). Then they saw the problem of having multiple stores laying around, looked around and figure out there is this nice architecture and implemented it in redux.

I think we as the full Drupal community should experience the same path for typesystems like typescript/flow. While I totally agree that they totally add huge amount of value on custom projects, I think the overall Drupal community isn't there to really use this tools. Without the experience over time, which custom projects have too, it is less obvious what these tools over. 
When we do though experience the path, it might become clear whether typescript of flow is the better tool for us. Once we are there then, we can actually use it properly, which means for example using https://flow.org/en/docs/types/unions/ as much as possible for example.

rsanzante:
I would like to add some thoughts to this conversation:

- While JS is very popular and TypeScript is not so popular, Angular is the second popular library/framework/tool, and Angular uses TypeScript.

[...]

PR #526, opened 2018-11-04
Typescript - The rough cut.

martinfrances107:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/64

This is a unreviewable mega patch

Experimental interfaces have been created .. all function/method signatures have been locked down with type information. [FWIW I have dumped many of the new interfaces in the constants directory.
]

It is a rough draft mega patch -- just so I can see how the whole things fits together before I make refinements ... From which I plan to make small targeted patches.

Here I just want to check  github and the running of tests.


packages/admin-ui/src/constants/message_severity.ts
```diff
@@ -0,0 +1,6 @@
+export enum MESSAGE_SEVERITY {
```
dawehner: I do like this, making impossible states not allowed

martinfrances107: Is there any way to split it up somehow into logical pieces?

That was always the plan ... I was also hoping  to have it as a reference if only to ease conversations where I can point at aspects of it - use it as a short hand for want we want to do ... or more likely - given new information say "this is what is looks like when I go wrong."

When I start make a list of obvious fragmentation ... These float to the top... which I intend to work on.

A) MESSAGE_SEVERITY -- issue

B) The ACTION_TYPE issue 
  ... although what I have done is moved CONSTANTS that were contained in separate modules and moved them into a single file  constants/action_type.ts  [ I have bunched the constants so that I keep the history of which file I moved them from and left comments like "see roles.ts"  ] But I struggle with this design choice ... should I keep the constant in the relevant module and make a mega ENUM using the union operator '|' .. Should I be using interface not ENUM ... I am still sitting on that. If they have to be unique and do not need to be iterable .. well that is the definition of a SET over MAP in javascript. hmm

C) The Node/Schema/Content issue....
     utils/api.ts   Has a "Node" interface ... which I think is a bodge. Hand crafted typescript which has to be keep in sync with changes to core.Agh  There are similarly troubling Schema and Content interfaces. But hey this is a rough draft. The correct thing todo is to autogenerate this for a core definition somehow...

D) FileUpload and FileUploadWidget... have too complex a set of data structures. I appreciate to some extent we have to receive radically different structures base on whether we are in single or multiple mode... but I am looking for a more elegant expression of that fixed.complexity.

E) Slowly convert each sub_atom and then widget in is small easy to review PR's

dawehner: B) It looks like you can iterate over ENUM's, see https://github.com/Microsoft/TypeScript/issues/17198#issuecomment-315400819
C) It feels like this could be its own issue where we first discuss how a good solution could look like, instead of trying to make a PR directly. What you describe as autogeneration might be tricky though. When the content model changes, we can't actually change our types, I guess.
D) I think FileUpload/FileUploadWidget might be up for some general refactoring. As part of that we could introduce typescript, as it would help to do that
E) Slowly convert each sub_atom and then widget in is small easy to review PR's

MrOUTSTANDING9:
Wanting to look into the matter as soon as possible

martinfrances107:
>Does this make sense? 

Yep I am happy to close.

I will keep the branch alive on my account so I can refer back to anything I want to talk about

https://github.com/martinfrances107/drupal-admin-ui/tree/typescript

Issue #527, opened 2018-11-05
Api.js - Fix restoreState() storedState - type confusion 


martinfrances107:
This is just a minor  type checking issue thrown up while I try and convert this issue to typescript

the variable **storedState** local to the method restoreState() is being used for 2 completely different purposes.

A) The **string** variable  input to a parse call 
B) The **object** variable,  the output of the parse call.

This is what I want the parse statement to look like 

storedState = JSON.parse(stringifiedState);

I am just breaking off a small aspect of the typescript branch and fixing it early

```
/**
 * Restore from local storage.
 */
const restoreState = () => {
  let stringifiedState: string = '';
  let storedState: object = {};
  if (typeof window === 'object') {
    try {
      // Test for Safari private browsing mode. This will throw an error if it can't set an item.
      localStorage.setItem('localStorageTest', true);
      stringifiedState = localStorage.getItem(localStorageName) || '';
    } catch (e) {
      // In case like Safari private browsng mode we don't support any restoring.
      // Also note: enzyme has window but no Cookie set.
      stringifiedState =
        (window.Cookie &&
          decodeURIComponent(window.Cookie.get(localStorageName))) ||
        '';
    }
  }

  try {
    storedState = JSON.parse(stringifiedState);
  } catch (e) {
    storedState = {};
  }
  return storedState;
};
```

martinfrances107:
Thank you.

PR #529, opened 2018-11-05
Stringified state

martinfrances107:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/527

This is just a minor issue 

Here is a summary of the originating issue.

As I work on the typescript branch -- the lifecycle of the storedState  variable is driving the linter crazy.

It is initialized as a object, assigned a string, and the converted back to a object as part of a JSON.parse() 

The solution is to split the variable into two.

### DIFF
--- a/packages/admin-ui/src/App.js
+++ b/packages/admin-ui/src/App.js
@@ -43,8 +43,8 @@ export const localStorageName = 'drupalAdminUiReduxState';
  * Restore from local storage.
  */
 const restoreState = () => {
-  const storedState;
-  const stringifiedState;
+  let storedState;
+  let stringifiedState;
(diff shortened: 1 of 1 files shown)


packages/admin-ui/src/App.js
```diff
@@ -43,24 +43,25 @@ export const localStorageName = 'drupalAdminUiReduxState';
  * Restore from local storage.
  */
 const restoreState = () => {
-  let storedState = {};
+  const storedState;
```
dawehner: Nice, more const is always nice!

dawehner:
@martinfrances107 Do you have an idea about those failures?

Issue #532, opened 2018-11-06
Installation fail.


martinfrances107:
I have just cleared away my drupal-admin-ui  directory structure and am having trouble reinstalling

after  composer clear-cache

I follow the readme  with this command and here is the resulting error messages.

```
composer create-project jsdrupal/drupal-admin-ui-demo -s dev --prefer-dist
Installing jsdrupal/drupal-admin-ui-demo (dev-master 3b7f86cb78e2d86af1ffb165938397e550c1feff)
  - Installing jsdrupal/drupal-admin-ui-demo (dev-master 3b7f86c): Downloading (100%)         
Created project in /Users/martin/sites/drupal-admin-ui-demo
Could not scan for classes inside "docroot/core/lib/Drupal.php" which does not appear to be a file nor a folder
Could not scan for classes inside "docroot/core/lib/Drupal/Component/Utility/Timer.php" which does not appear to be a file nor a folder
Could not scan for classes inside "docroot/core/lib/Drupal/Component/Utility/Unicode.php" which does not appear to be a file nor a folder
Could not scan for classes inside "docroot/core/lib/Drupal/Core/Database/Database.php" which does not appear to be a file nor a folder
Could not scan for classes inside "docroot/core/lib/Drupal/Core/DrupalKernel.php" which does not appear to be a file nor a folder
Could not scan for classes inside "docroot/core/lib/Drupal/Core/DrupalKernelInterface.php" which does not appear to be a file nor a folder
Could not scan for classes inside "docroot/core/lib/Drupal/Core/Site/Settings.php" which does not appear to be a file nor a folder
Gathering patches for root package.
> DrupalProject\composer\ScriptHandler::checkComposerVersion
Loading composer repositories with package information
Installing dependencies (including require-dev) from lock file
Nothing to install or update
Generating autoload files

                                                                                                                   
  [RuntimeException]                                                                                               
  Could not scan for classes inside "docroot/core/lib/Drupal.php" which does not appear to be a file nor a folder  
                                                                                                                   

create-project [-s|--stability STABILITY] [--prefer-source] [--prefer-dist] [--repository REPOSITORY] [--repository-url REPOSITORY-URL] [--dev] [--no-dev] [--no-custom-installers] [--no-scripts] [--no-progress] [--no-secure-http] [--keep-vcs] [--remove-vcs] [--no-install] [--ignore-platform-reqs] [--] [<package>] [<directory>] [<version>]
```

[...]

PR #534, opened 2018-11-08
Greenkeeper/initial

martinfrances107:
#528  The last PR was out of date so I performed the following

1) I rebased greenkeeper/initial to pick up the latest master changes.
2) I run "yarn install" to refresh the packages... see the changes to   yarn.lock 

The module has a critical "installation fails" issue (#532) .. I suggest this issue must be resolved before we can work on that ... so that makes this issue critical.


martinfrances107:
Is there a reason the state changes are included on this PR?

Does that disrupt commit workflow?

It is a consequence of the fact that I did a reroll - as step one.

The old broken PR -  is behind by one commit the one that added state changes.

I didn't have permission to auto-update.

mattgrill:
@martinfrances107 - I think it makes sense to scope this PR to either be about GreenKeeper updates or fixes to the stored state in `packages/admin-ui/src/App.js`

martinfrances107:
It is a question of what the diff is differencing 

The tests infrasture is  doing a diff from master and reporting. "There are not conflicts with the base branch".

When you click on the files changes tab ,,,

The diff is differencing between two branches both labelled "greenkeeper/initial" but from two different users.

So I think the state changes which are already committed to master are just an artifacts.

PR #535, opened 2018-11-10
[#532] Fixes the demo install

justafish:
## Issue


justafish:
Schemata stopped wrapping things in a `data` object

dawehner:
I'll try to fix the remaining problems.

.circleci/config.yml
```diff
@@ -160,11 +159,6 @@ jobs:
     steps:
     - fix-php-env
     - checkout
-    - run:
```
dawehner: Do you mind explaining why we stop generating the lock file? ```update --lock``` shouldn't update the core version at least.

dawehner: I thought we have that so we can increase version numbers inside the `admin_ui_support` module.

PR #494, opened 2018-09-22
Add route override for entity.node.edit_form

alexburrows:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/493

## Screenshot / UI changes

When editing node from node view it now routes to correct vfancy route

## Testing instructions

View an existing node and then click edit, it will take you to the Drupal 8.x node edit form. Then apply pull request and it will redirect to vfancy entity.edit_form

PR #536, opened 2018-11-10
Copy of https://github.com/jsdrupal/drupal-admin-ui/pull/494 with tests

dawehner:


admin_ui_support/tests/src/Functional/RedirectRoutesTest.php
```diff
@@ -40,15 +52,22 @@ public function testRouteRedirect() {
...
     $route_provider = \Drupal::service('router.route_provider');
     foreach ($paths as $path) {
       $routes = $route_provider->getRoutesByPattern($path);
-      $this->assertEquals(1, $routes->count());
-      foreach ($routes as $route) {
-        $this->assertEquals('Drupal\admin_ui_support\Controller\DefaultController::getAppRoute', $route->getDefault('_controller'));
+      $this->assertGreaterThanOrEqual(1, $routes->count());
+      foreach ($routes as $route_name => $route) {
+        // The canonical node route /node/1 is a route suggestion for
+        // /node/1/edit.
```
tedbow: Should we just test for this in an `else`

dawehner: What do you mean with that. Have a special test for the `/node/` routes?

PR #537, opened 2018-11-11
[#64] Adds TypeScript support and converts one simple component

justafish:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/64

### DIFF
--- a/packages/admin-ui/src/NoMatch.tsx
+++ b/packages/admin-ui/src/NoMatch.tsx
@@ -1,19 +1,18 @@
 import { Component } from 'react';
-import PropTypes from 'prop-types';
 
 import routes from './routes';
 
-const NoMatch = class NoMatch extends Component {
-  static propTypes = {
-    location: PropTypes.shape({
-      pathname: PropTypes.string.isRequired,
-    }).isRequired,
-  };
+export interface Props {
+  location: {
+    [pathname: string]: string,
+  },
+}
 
+class NoMatch extends Component<Props> {
   componentDidMount() {
     // We want to redirect everything without a match back to Drupal.
     if (!Object.keys(routes).includes(this.props.location.pathname)) {
-      window.location =
+      window.location.href =
         process.env.REACT_APP_DRUPAL_BASE_URL + this.props.location.pathname;
     }
   }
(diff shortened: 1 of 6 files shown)


packages/admin-ui/tsconfig.json
```diff
@@ -0,0 +1,27 @@
...
+    "resolveJsonModule": true,
+    "isolatedModules": true,
+    "noEmit": true,
+    "jsx": "preserve"
+  },
+  "include": [
+    "src"
+  ],
+  "exclude": [
+    "src/tests"
```
dawehner: Why would we exclude tests?

packages/admin-ui/src/NoMatch.tsx
```diff
@@ -1,19 +1,18 @@
...
+    [pathname: string]: string,
+  },
+}
 
+class NoMatch extends Component<Props> {
   componentDidMount() {
     // We want to redirect everything without a match back to Drupal.
     if (!Object.keys(routes).includes(this.props.location.pathname)) {
-      window.location =
+      window.location.href =
```
dawehner: Isn't that doing the same?

packages/admin-ui/tsconfig.json
justafish: The nightwatch tests run on node, and does things like require-ing babel-register, which causes CRA to throw errors

packages/admin-ui/src/NoMatch.tsx
justafish: Apparently not https://stackoverflow.com/questions/13106950/set-window-location-with-typescript

dawehner: Oh I see, this is TS speciality

dawehner:
> I have played with this PR and it looks good... in a moment of paranoia I checked that the build size was unchanged. That is true but as with the typescript monster branch ... the build time increases significantly for me almost doubles.

As long its the production build time, that's totally fine IMHO.

packages/admin-ui/src/NoMatch.tsx
dawehner: Thank you

packages/admin-ui/tsconfig.json
dawehner: I'm wondering what this mean for unit tests, can they call their appropriate TS alternatives? Anyway, we can always iterate on that

dawehner:
Thank you @martinfrances107 for giving a +1

PR #540, opened 2018-11-11
Use typescript for fieldItem.ts

dawehner:


martinfrances107:
This is good.

A) It tightly focused -- and will be a benefit to the next developer adapting the code.
B) In comparison to my  typescript mega patch .... this code fragment is better as it the fieldItem interface locks the code down more.

packages/admin-ui/tsconfig.json
```diff
@@ -1,8 +1,9 @@
 {
   "compilerOptions": {
     "lib": [
-      "ES7",
-      "DOM"
+      "es7",
+      "es2017",
+      "dom"
```
justafish: why did this need changing? I took the values from https://www.typescriptlang.org/docs/handbook/compiler-options.html

dawehner: Sorry I should have commented. Es2017 is needed for ```Object.values```.

dawehner: For the lowercase versions ... webstorm complained

Issue #539, opened 2018-11-11
Linting: Check ts and tsx files. 


martinfrances107:
We are dipping our toes  into typescript ... which makes me happy.

I just want to iron out a little kink.

yarn run test:lint 

after inspecting 

 reports/eslint-junit.xml

It is clear that only js files are inspected, even if  ts files are present

I am going to cut and paste the relevant config from the typescript branch and make a PR.

dawehner:
Thank you!

martinfrances107:
I am posting early, 

I just want to create a incomplete PR just so some of the refinements that come later are obvious in the review process.

Things like single-quote and double-quote issues MAYBE should be autofixable.

martinfrances107:
I am trying to move the project in a positive direction, but what I am doing maybe controversial -- so I best highlight it ( see B) 

A) I have moved tsconfig.json and tslint.json up into the top level directory ... so it acts on every project just like .eslintrc.json did.

B) This is the controversial bit, because I extend "js:recommeded" see tslint.json
we pick up a few rules which spew out lots of failures. In the first version I am turing off the following rules from that standard

      "trailing-comma": false,
+      "semicolon": false,
+      "quotemark": false,
+      "object-literal-sort-keys": false,
+      "no-shadowed-variable": false,
+      "max-line-length": false,
+      "no-empty": false,

My line is that with a autofixer ... it is trivial to fix these issues in our design. ( shadowed-variable ) will take some manual care to fix.  

If we are in agreement I will file new issue and let the autofix blow away our troubles. ( please think autumn and leaf blowers I am not a wannabe gangster. )

martinfrances107:
OK I will sleep on it...  small changes are always good.

martinfrances107:
After a little nudge from justafish 

I am dropping the old PR -- instead of the pulling in tslint packages, it is simpler to get eslint to parse .ts and .tsx files.

just to be clear, one little non-obvious kink is that .eslintrc files cannot as yet be configured to parse typescript files instead it MUST be specified on the command line

Also to smooth the review I am breaking the PR into two stages. When I turn on typescript linting - It highlights errors. I will let testbot fail and then I will fix them in a second push to the PR

> Specifying File extensions to Lint
> Currently the sole method for telling ESLint which file extensions to lint is by specifying a comma separated list of extensions using the --ext command line option. Note this flag only takes effect in conjunction with directories, and will be ignored if used with filenames or glob patterns.

[...]

PR #541, opened 2018-11-11
Add editor config file

dawehner:


martinfrances107:
Just to bind two issues together

As typescript is becoming a thing.  I think we are about to make use of autofixing 
I have been thinking about this as part of Linting: Check ts and tsx file ( https://github.com/jsdrupal/drupal-admin-ui/issues/539 ) 

Where the existing rules coded into .eslintrc - are being worked into a yet to be created  tslint.json and tsconfig.json. Where the rules I was thinking about extending  a standard ruleset for example "tslint:recommeded" and "tslint-config-prettier"

I am happy to see this PR be merged 

I just want to say in public is that we/I may have some reconciliation work, after the fact,   so that the coding standard rules seen by the IDE ( those affected by this config file ) and those imposed by the combination of tsconfig.json and tslint.json all work together.

In short this is ready to be merged.

PR #542, opened 2018-11-12
Ts linting

martinfrances107:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/539

This is a early version of the PR. I want the failures to be public ... so the resolution is clear.


martinfrances107:
When I moved tsconfig.json and tslint.json -- that was a mistake. 

The build process misses the files and inserts default versions of those files.

martinfrances107:
After a chat with justfish,

I am closing this PR and starting again to keep things clean.

tslint is not a package we want to pull in. We can get the linting we want using eslint.

PR #551, opened 2018-11-19
esLinting: Check ts and tsx files. #539

martinfrances107:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/539

The problem this PR solves is that ts and tsx files do not show up in the eslint reports

think 

yarn run test:lint 

That argument also applied to 

yarn run eslint

This PR has errors - I think it is an important step to demonstrate publically the new errors are highlighted by the fix.

In a second push to this PR I will show the errors being fixed.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -4,6 +4,10 @@
     "plugin:prettier/recommended",
     "plugin:jsx-a11y/recommended"
   ],
+  "parser": "typescript-eslint-parser",
+  "parserOptions": {
+    "jsx": true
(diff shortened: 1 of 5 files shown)


martinfrances107:
Above I was expecting detailed linting issues. ... I guess the error reporting I was expecting was based on previous test fails.

lauriii:
Thank you!

martinfrances107:
@lauriii  Thanks for following along and the quick review

lauriii++

Issue #546, opened 2018-11-16
Storybook is broken


lauriii:
Storybook seems to be currently broken. It seems like there's multiple PR's that has caused the regression.

We have to move `.storybook` folder into the package folder (https://github.com/jsdrupal/drupal-admin-ui/pull/516) and we have to make the storybook config support TypeScript (https://github.com/jsdrupal/drupal-admin-ui/pull/537)

martinfrances107:
I can confirm this bug.... I will take a look at this tomorrow

I think .storybook does needs to be copied from the root directory 
down into 

packages/admin-ui

After that 

yarn workspace admin-ui storybook

does not completely work.. but this time the server starts ..  it needs further work

here is the next stacktrace 

`    ReferenceError: exports is not defined
    at Module../src/components/02_atoms/FileUpload/FileUpload.js (http://localhost:9001/static/iframe.bundle.js:91027:23)
    at __webpack_require__ (http://localhost:9001/static/iframe.bundle.js:725:30)
    at fn (http://localhost:9001/static/iframe.bundle.js:102:20)
    at Module.<anonymous> (http://localhost:9001/static/iframe.bundle.js:91729:19)
    at Module../src/components/02_atoms/FileUpload/FileUpload.stories.js (http://localhost:9001/static/iframe.bundle.js:91765:30)
    at __webpack_require__ (http://localhost:9001/static/iframe.bundle.js:725:30)
    at fn (http://localhost:9001/static/iframe.bundle.js:102:20)
    at webpackContext (http://localhost:9001/static/iframe.bundle.js:90994:9)
    at http://localhost:9001/static/iframe.bundle.js:90967:12
    at Array.forEach (<anonymous>)`

martinfrances107:
I am just posting an initial version....

Here is a comment which points out where the error is and what I don't understand

"exports is coming from commonJS"

https://stackoverflow.com/questions/43042889/typescript-referenceerror-exports-is-not-defined

martinfrances107:
This is the first unmet dependency we have 

> warning "workspace-aggregator-1b2a520b-178a-4eb0-af00-57875c464673 > admin-ui > @storybook/react@4.0.4" has unmet peer dependency "babel-loader@^7.0.0 || ^8.0.0".

So I am picking out quotes from a relevant open storybook issue.

https://github.com/storybooks/storybook/issues/4764

[...]

PR #554, opened 2018-11-22
Storybook files with JSX elements need the correct file extension

martinfrances107:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/546

This is only a partial fix for the broken story book issue.

files containing JSX elements must have the file extension jsx 

recent changes enforce that rule and hence storybook is broken.


martinfrances107:
This is a question about linting standards looking at the master branch of drupal-admin-ui the question is the entry in .eslintrc.json look broken, but is maybe deliberate  it

jsx elements are allowed in .js files ... this seems wrong but to correct that would be a PR which changes 41 file extensions.. 

I am leaving that alone and just allowing jsx elements in jsx files. which is cause the current lint problems.

dawehner:
@martinfrances107 Thank you!

PR #517, opened 2018-10-27
Experiment with adding dynamic components from modules

dawehner:
https://github.com/jsdrupal/drupal-admin-ui/issues/476

PR #560, opened 2018-11-26
Extension points for module provided routes

mattgrill:
Extension points for Drupal module provided routes & components. This process uses the same loading component and bundling module developed in #517.

- Before the `<App />` is rendered to the page, a request to Drupal is made to ask for any available routes.
- Routes are merged with routes provided by `drupal-admin-ui`.
- Each route provided by Drupal is prefixed with the module that provided it.


packages/extension-points/README.md
```diff
@@ -35,3 +35,14 @@ widgets:
     component: js/build/options_buttons.widget.js
     multiple: true
 ```
+
+### Routes Configuration
+
+A single `*.admin_ui.routes.yml` file can contain as many routes as needed. Note that route will be prefixed with the key name. For example the following route would be accessible at, `/admin_ui_widget_example/test`.
```
lauriii: What is the key name referring to? Do you mean module name?

mattgrill: I do.

packages/extension-points/README.md
```diff
@@ -35,3 +35,14 @@ widgets:
     component: js/build/options_buttons.widget.js
     multiple: true
 ```
+
+### Routes Configuration
```
dawehner: Should we document that for  menu links they can still provide them as normal menu link inside a drupal module?

packages/admin-ui/src/index.js
```diff
@@ -11,5 +12,10 @@ if (process.env.NODE_ENV !== 'production' && process.env.REACT_APP_AXE) {
   axe(React, ReactDOM);
 }
 
-ReactDOM.render(<App />, document.getElementById('root'));
+api('admin_ui_routes').then(({ routes }) => {
```
dawehner: do you think its the best idea to do this that early in process? Can't we start to bootup the application  already and then  do the http request? I'm just curious what your thoughts are.

packages/extension-points/index.js
```diff
@@ -16,21 +16,30 @@ Promise.all([
       operatingDirectory,
     )}.admin_ui.components.yml`,
   ),
+  readFilePM(
```
dawehner: What's that?

packages/extension-points/README.md
mattgrill: Do we have any documentation about menu links already?

packages/admin-ui/src/index.js
mattgrill: I was trying to think about where to load these routes. If they are loaded as early as possible, there shouldn't be any delay (fetching additional available routes), if a user tries to navigate to the page that is part of a module provided route.

packages/extension-points/index.js
mattgrill: `const [readFilePM] = [readFile].map(promisify);`

packages/extension-points/README.md
```diff
@@ -35,3 +35,14 @@ widgets:
...
+
+A single `*.admin_ui.routes.yml` file can contain as many routes as needed. Note that route will be prefixed with the Drupal module name. For example the following route would be accessible at, `/admin_ui_widget_example/test`.
```
lauriii: Makes more sense now. Do you think it would make sense to add a mention that the following code is located in `admin_ui_widget_example` module?

mattgrill: 👍

packages/admin-ui/src/index.js
lauriii: I think it's hard to make any assumption at this point what problems we might run into with this. Maybe we should just go with the current solution and optimize it later based on feedback?

dawehner: 👍

Issue #564, opened 2018-12-01
Use a SPDX license identifier 


martinfrances107:
The license identifier we use is deprecated

This is a parallel issue to 

https://www.drupal.org/node/2935137

License "GPL-2.0+" is a deprecated SPDX license identifier, use "GPL-2.0+-only" or "GPL-2.0+-or-later" instead

To generate my PR I am visiting each composer.json file  - and will run

composer validate 

That makes the problem visible.

Of the two possible new identifiers I plan to use "GPL-2.0+-or-later"

martinfrances107:
Thank you.

PR #565, opened 2018-12-01
SPDX license identifier #564

martinfrances107:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/564

run 

cd to admin_ui_support
composer validate 

see that by this PR one of the error/warnings  is cleaned up.


dawehner:
This makes sense 100%

Issue #317, opened 2018-08-11
Minimum PHP version in README file


progga:
Hello,
I was following the instructions in the README to setup drupal-admin-ui locally.  I got told off by composer almost as soon as I got started :(  The root of the trouble is that I am on Debian Stable (AKA Stretch) which is carrying PHP 7.0.x and [doctrine/annotations](https://packagist.org/packages/doctrine/annotations) et al now demands **php ^7.1**.  The otherwise excellent README mentions **PHP 5.5.9 or greater** as a requirement and this seem to have become outdated with the advent of Drupal 8.7.x.  So I thought I will update the README accordingly.  Pull request to follow.

As an aside, I had to summon the help of Docker to get me out of trouble.  After tinkering around for a day or so, I eventually managed to reach the promised land.  I have left the [Dockerfile as a gist](https://gist.github.com/progga/37bb352c18af46ba2489fed6414e1b15) in case anyone finds it useful.

dawehner:
I think we should limit the version of the doctrine library to be as stricter. IMHO following core's PHP version makes sense. What do you think about this alternative proposal?

progga:
Hello @dawehner 

> I think we should limit the version of the doctrine library to be as stricter.

If we want to take this route, we can **drop** jsdrupal/drupal-admin-ui-demo's composer.lock file.  [drupal-composer/drupal-project](https://github.com/drupal-composer/drupal-project) seem to have fathered jsdrupal/drupal-admin-ui-demo and that one does **not** commit its composer.lock file.  Perhaps we shouldn't either.  When composer.lock is absent, composer will install whatever version is appropriate.

> IMHO following core's PHP version makes sense.

Totally makes sense :)

## See also
- https://www.drupal.org/project/drupal/issues/2843328

dawehner:
Well, an alternative would be setting the requirement on doctrine to ensure this doesn't happen.

I think especially for having a stable environment for testing etc. having a composer.lock is useful.

Note: drupal project doesn't ship with one because they are just a template for your own site.

progga:
> an alternative would be setting the requirement on doctrine to ensure this doesn't happen.

Okay, to go down this way we will have to pin the following 12 packages to their corresponding versions from [drupal/drupal's composer.lock](https://github.com/drupal/drupal/blob/8.6.x/composer.lock):
- doctrine/annotations: ~1.2.7
- doctrine/cache: ~1.6.1
- doctrine/collections: ~1.3.0
[...]

PR #566, opened 2018-12-03
We need at least PHP 7.0.8 for drupal-admin-ui-demo

progga:
Note: PHP 7.0.8 is the minimum PHP 7.x requirement for drupal/core

## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/317


lauriii:
Should we add this requirement to the composer.json file as well?

justafish:
Thank you!

PR #567, opened 2018-12-03
correct openAPI dependencies

martinfrances107:
## Issue

This is not a complete fix for the installation failure... I am just cleaning something up while I see it.

There are several critical errors hiding behind the another.

While I was looking for a solution to one 

>Unable to install modules jsonapi, admin_ui_support due to missing modules   
  admin_ui_support. 

I looked over the validity of two files in admin-ui-support

composer.json and admin_ui_support.info.yml

and I found this dependency failure 

openapi/openapi is listed when we need to say druapl/openapi

if we intend to pull in 

https://www.drupal.org/project/openapi


martinfrances107:
@lauriii thanks for the update... that is useful information.

I am still try to get to the bottom of this error ... but this PR has not future.

PR #570, opened 2018-12-03
Make it possible to have a missing .router.yml file

dawehner:


packages/extension-points/package.json
```diff
@@ -1,11 +1,14 @@
 {
-  "name": "@drupal/extension-points-bundler",
+  "name": "extension-points",
```
dawehner: If  we don't have that, ```yarn workspaces ... prettier``` doesn't work.

mattgrill: What happens when we want to publish this to npm? Also we are dropping the `bundler` part of this project name?

dawehner: Good point, I'll  revert it

PR #577, opened 2018-12-07
Run composer update with --lock flag

lauriii:
## Issue

## Description
It seems some changes to the build process caused [vendor files being added to the dist build](https://github.com/jsdrupal/drupal-admin-ui-demo/commit/44cbea61de116db8edf8054ef8751be125a64b0d) which broke `composer create-project`.


martinfrances107:
This is the solution I was looking for but could not find.

laurii++

Thank you

dawehner:
Perfect!

PR #579, opened 2018-12-10
Optimize @drupal/admin-ui-utilities build process

mattgrill:
## Issue

The `@drupal/admin-ui-utilities` package is quite large, `26.6 KiB`.

The majority of this bundle size is the regenerator runtime added from `@babel/plugin-transform-runtime` and that our `@babel/preset-env` configuration specifies only `last 2 versions`. 

Browser list has `last 2 versions` and providing **85.25%** global browser coverage.

## Resolution

Removing `@babel/plugin-transform-runtime` and with it the regenerator runtime reduces the bundle size to a more manageable `14.5 KiB`. However, `last 2 versions` includes many versions of browsers we probably don't need to care about. Using a more reasonable set of browsers, `ios >= 10.3, chrome >= 55, firefox >= 53, safari >= 10.3, ChromeAndroid >= 70, edge >= 15, opera >= 42` keeps our coverage at **80.3%**.

## TL;DR

- Bundle size from `26.6 KiB` to `14.5 KiB`
- Browser global coverage from **85.25%** to **80.3%**
	- This is mostly due to the removal of support for things like `Android Browser`, `Blackberry Browser`, `IE Mobile`, `Samsung Internet`, and `IE 10/11`


lauriii:
I'm wondering if we should make this change based on the proposal in here: https://www.drupal.org/project/drupal/issues/1030090?

dawehner:
I'm wondering whether we could move the babel config into the root folder, so its synched with https://github.com/jsdrupal/drupal-admin-ui/blob/master/packages/admin-ui/package.json#L6.

mattgrill:
@dawehner Is this just for node targets?

dawehner:
> Browser list has last 2 versions and providing 85.25% global browser coverage.

That's quite a low number, but maybe that's just me. It feels like supporting similar IE versions as Drupal core seems a good path to follow.

PR #580, opened 2018-12-11
Adds a Docker setup for this project

justafish:
## Issue
Setup is hard!

## Testing instructions
- Checkout this branch and ensure it's free of any previous build files (e.g. a clean checkout or `git clean -f -d -X`)
- Follow the new instructions in the README
- Make sure you can run Nightwatch tests
- Profit!

PR #581, opened 2018-12-12
add a nvmrc

dawehner:
Everytime I try to work on this project my current node version is potentially wrong.
NVM solves this problem by managing your node versions

Note: This is useful, even beside of https://github.com/jsdrupal/drupal-admin-ui/pull/580 for things like `audit, add` etc. People might run them outside of the container.


justafish:
@dawehner yarn already throws an error if you try to run anything with the wrong version of node, is this not happening for you?

justafish:
which is taken from https://github.com/jsdrupal/drupal-admin-ui/blob/master/package.json#L8

dawehner:
@justafish Well that's the point, with a nvmrc in place you don't need to get the error in the first place.

dawehner:
Well, other's seems to not have these problems ¯\_(ツ)_/¯

Issue #555, opened 2018-11-22
JSX elements are not being linted fully


martinfrances107:
This maybe a deliberate design choice that I am ignorant of 

BUT 

we are getting compile errors in relation to storybook when we have JSX elements inside js files
[ that is being fixed here #554 as part of #546 ]

This issue is separate I just want clearance to lock this down so that all files using JSX elements have the extension jsx or tsx.

dawehner:
It looks like there is a disagreement between airbnb and people working on react itself, see https://github.com/facebook/create-react-app/issues/87#issuecomment-234627904

Personally I think I agree with @gaeron's perspective.

martinfrances107:
I am happy to follow the no jsx files convention

BUT I am seeing some horrible interaction with our current linting configuration files

case in point if you rename AddContent.test.js to  AddContent.test.jsx 

then you get **new** linting errors 

> Warning: validateDOMNesting(...): <div> cannot appear as a descendant of \<p\>.

I am going to look into this more so that we get the same error warning from js files with JSX elements.

dawehner:
> I am going to look into this more so that we get the same error warning from js files with JSX elements.

dawehner:
>  I am going to look into this more so that we get the same error warning from js files with JSX elements.

Nice, thank you!

martinfrances107:
Great - the loose ends are complete and merged.. 

thank you.

PR #584, opened 2018-12-17
AddContent: Replace div with fragment

martinfrances107:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/555

The parent issue identifier that typescript files were not being linted. Another PR rename files containing typescript to AddContent.test.js and  AddContent.test.jsx 

This PR now fixes one of the outstanding lint errors 

Markup is wrapped in a ListItemText element and so as we sanitize it is inappropriate to use Markup's default of a div as an internal wrapper.

https://github.com/jsdrupal/drupal-admin-ui/issues/555#issuecomment-441188250

Here is the lint warning

> Warning: validateDOMNesting(...): cannot appear as a descendant of <p>.

looking at the valid values for tagName in  Markup -- "span" or "fragment" are the only possibilities 

fragment has the advantage of dissolving into nothing.

### DIFF
--- a/packages/admin-ui/src/components/05_pages/AddContent/AddContent.js
+++ b/packages/admin-ui/src/components/05_pages/AddContent/AddContent.js
@@ -47,6 +47,7 @@ export default class extends Component {
                   secondary={
                     <Markup
                       content={this.props.contentTypes[contentType].description}
+                      tagName="fragment"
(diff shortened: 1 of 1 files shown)


martinfrances107:
Thank you both

PR #589, opened 2018-12-20
Handle missing broken routes endpoint

dawehner:
When the routes endpoint fails the entire application is broken.
Let's protect against that. I'm not sure this approach is the best one, but at least its something.

### DIFF
--- a/packages/admin-ui/src/index.js
+++ b/packages/admin-ui/src/index.js
@@ -25,4 +25,4 @@ api('admin_ui_routes')
       document.getElementById('root'),
     );
   });
- registerServiceWorker();
+registerServiceWorker();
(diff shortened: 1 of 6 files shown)


packages/admin-ui/src/index.js
```diff
@@ -14,10 +14,15 @@ if (process.env.NODE_ENV !== 'production' && process.env.REACT_APP_AXE) {
...
-api('admin_ui_routes').then(({ routes }) => {
-  ReactDOM.render(
-    <App serverRoutes={routes} />,
-    document.getElementById('root'),
-  );
-});
-registerServiceWorker();
+api('admin_ui_routes')
+  .catch(err => {
+    console.err(err); // eslint-disable-line no-console
```
mattgrill: Would it be possible to use the error message interface to present something to the user? I don't think this is critical though.

dawehner: @mattgrill I guess we could initialize the application with an error?

mattgrill: Lets make a followup for that. Probably some more involved changes there.

dawehner: Sure: https://github.com/jsdrupal/drupal-admin-ui/issues/600 is the followup.

Issue #595, opened 2018-12-21
One process per container.


martinfrances107:
There is something I want to discuss.... My mind is not certain - so I would like to hear what others have to say. I like the recent effort to move to docker - and in what I say next there is no criticism.

Let me start by referencing the install modules 

apk add php sqlite composer git \
    php7-pdo php7-sqlite3 php7-pdo_sqlite php7-gd php7-openssl php7-json \
    php7-mbstring php7-dom php7-session php7-xml php7-simplexml php7-tokenizer \
    php7-curl php7-xmlwriter php7-ctype php7-opcache php7-pdo_mysql mysql-client \
    php-apache2 apache2 apache2-ctl

There is a mantra in docker of "one process per container". We have php, mysql and apache

So the "high availability" perspective is : - 

When something goes wrong I want to limit my actions to reprovisioning a single failed container. In the one process per container model the innards of the container will be small so I can rapidly tear down and bring up that "service"  - In the current configuration if the apache process fails what is wrong is that I then I find I have to restart the database.

From the perspective of a security review 

The advantage of one process per container is that the directory structure can be restricted to the single task at hand. In the current configuration IF a hacker gets into the apache container then the existence of volumes mounted in the container related to mysql and php become a problem-- In theory they permit direct access to the database and the php volumes allow the initialization of php in server mode which allows for the construction of a black hat parallel server.

So now for the arguments on the other side of the scales. 

Let me start by saying I initially reported problems saying that the performance on my old mac was bad - adjustment were rapidly made which mitigated the problem -  but performance is still not great.

The suggested reconfiguration adds two extra containers so that apache, php and mysql can exist in isolation. This will further degrade performance.

**In short the current configuration is good for the lone developer working away on a single laptop...
to what extent should we support production concerns.**

My response was to buy a new laptop with 4 times the memory - but that is out of recognition that what I had has for a while not been fit for purpose.( especially in the light of docker )  So I discount my own perspective and I want to hear what other say.

justafish:
I'm a bit confused - MySQL and Apache are running in separate containers. Here are the list of services:
https://github.com/jsdrupal/drupal-admin-ui/blob/master/docker-compose.yml
I'm not sure what you mean by PHP and Apache being in separate containers?

> to what extent should we support production concerns
There is no intention for these containers to be used in production!

dawehner:
I totally agree with @justafish these containers are not meant to be used on production. In production you should not just blindly using someone else's container anyway.

[...]

PR #601, opened 2018-12-27
[#595] Remove SQLite from the Drupal Dockerfile

justafish:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/595


dawehner:
It looks like circle CI has an invalid cache entry somehow.

Issue #594, opened 2018-12-21
test_node fails frequently on CircleCI


justafish:
![nodeeditformrestoring_failed_dec-21-2018-091232-gmt 0000- coordinated-universal 1](https://user-images.githubusercontent.com/345613/50335058-0eed0a80-0502-11e9-9049-d99c3929ae78.png)

```
Timed out while waiting for element <[data-nightwatch="Edit Super easy vegetarian pasta bake"]> to be present for 5000 milliseconds. - expected [0;32m"visible"[0m but got: [0;31m"not found"[0m
    at Object.nodeEditFormRestoring (/root/drupal-admin-ui/packages/admin-ui/src/tests/Nightwatch/Tests/NodeEdit.js:36:13)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

dawehner:
There is a second random failure:

```
Timed out while waiting for element <[data-nightwatch="content-type-select"]> to be present for 5000 milliseconds. - expected [0;32m"visible"[0m but got: [0;31m"not found"[0m
```

dawehner:
Example of it failing:


https://circleci.com/gh/jsdrupal/drupal-admin-ui/7617?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

I've identified the potential error on the content page, but I wasn't yet successful for the node edit form.
https://github.com/jsdrupal/drupal-admin-ui/pull/603/files contains these fixes plus some additional logging.

PR #603, opened 2018-12-28
Some additional work to catch errors

dawehner:
https://github.com/jsdrupal/drupal-admin-ui/issues/594

### DIFF
--- a/packages/admin-ui/src/components/05_pages/Content/Content.js
+++ b/packages/admin-ui/src/components/05_pages/Content/Content.js
@@ -169,10 +169,11 @@ class Content extends Component {
       <PageTitle>Content</PageTitle>
       <LoadingBar style={{ position: 'relative', marginBottom: '5px' }} />
       <Paper>
-        {this.props.contentTypes && this.props.actions && (
+        {this.props.contentTypes && this.props.actions.length && (
           <Fragment>
             <div className={styles.filters}>
-              {this.props.contentTypes && this.props.actions && (
+              {/* @TODO Get rid of duplicated condition */}
+              {this.props.contentTypes && this.props.actions.length && (
                 <Fragment>
                   <TextField
                     inputProps={{ 'aria-label': 'Title' }}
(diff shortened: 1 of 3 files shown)


packages/admin-ui/src/components/05_pages/Content/Content.js
```diff
@@ -169,10 +169,11 @@ class Content extends Component {
       <PageTitle>Content</PageTitle>
       <LoadingBar style={{ position: 'relative', marginBottom: '5px' }} />
       <Paper>
-        {this.props.contentTypes && this.props.actions && (
+        {this.props.contentTypes && this.props.actions.length && (
           <Fragment>
             <div className={styles.filters}>
-              {this.props.contentTypes && this.props.actions && (
+              {/* @TODO Get rid of duplicated condition */}
+              {this.props.contentTypes && this.props.actions.length && (
```
dawehner: Given that this.props.actions is an array, treating it as a boolean doesn't work, as it gets evaluated to true.

dawehner: This could fix one of the source of failures.

packages/admin-ui/src/components/05_pages/NodeForm/index.js
```diff
@@ -49,13 +49,21 @@ class NodeForm extends React.Component {
     setErrorMessage: PropTypes.func.isRequired,
     onChange: PropTypes.func,
     requestComponentList: PropTypes.func.isRequired,
+    components: PropTypes.shape({
```
dawehner: The error is probably on the area of widgets, which weren't yet documented as proptypes.

packages/admin-ui/src/components/06_wrappers/ErrorBoundary/ErrorBoundary.js
```diff
@@ -15,6 +15,7 @@ class ErrorBoundary extends React.Component {
   }
 
   componentDidCatch(error) {
+    console.error('uncaught error', error); // eslint-disable-line no-console
```
dawehner: Logging the caught error should help to identify the potential source of error.

PR #605, opened 2018-12-28
Fix POST requests from api.js

dawehner:
Moving api.js to the utilities package introduced a couple of bugs,
one was saving no longer works.

As part of that I started to gitignore `build/index.js` as we don't need this in version control.


mattgrill:
@dawehner So I think this works. We should probably restore the build dir for now, and then figure out how we can actually deploy these modules to npm.

dawehner:
@mattgrill well, we would need to rebuild at least once, as otherwise this wouldn't be fixed. In general it feels like ```npm run  start``` should build this package as well somehow. Otherwise the DX is quite bad :|

mattgrill:
@dawehner Can we run this with a post-install script?

PR #612, opened 2019-01-13
Run composer test as the current user

dawehner:
When you use `docker exec` or `docker run` it executes the command as a root user. Therefore add the parameters to the composer test call as otherwise it fails.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -100,7 +100,7 @@ jobs:
       - start-project
       - run:
           name: Run tests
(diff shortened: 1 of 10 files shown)


dawehner:
@justafish Your branch doesn't appear for me to make a PR against. You could pull in https://github.com/jsdrupal/drupal-admin-ui/pull/612/commits/c5e6d88f562aee19568a7faadc0544cfb935c1ec and https://github.com/jsdrupal/drupal-admin-ui/pull/612/commits/a08450dea94ec3b05e421b2e32f037655e40c734 individually

Issue #508, opened 2018-10-23
Message() convert div into a text bearing element


martinfrances107:
InlineMessages.js

The following code  needs a little refactor

div is a logical container for other html elements
There are on 2 only text bearing elements in HTML SPAN and P

If I remember the css/html books I read a decade ago what happens under the hood is comical

`<div>The little house on the prairie</div>`

each white spaced element gets converted into its own html element -- so this is the resultant DOM representation

```
<div>
 <text>The</text>
  <text>little</text>
  <text>house</tex>
   ...
</div>


```

```
const Message = ({ message, messageSeverity }) => (
  <div className={`${styles.message} ${modifierClassMap[messageSeverity]}`}>
    {message}
  </div>
);
```

I will sort this issue out ... but it is low priority so it may happen slowly .. or maybe it is a good issue for a first timer... I am easy.

theyogendraDanwar:
Created pr for this issue: https://github.com/jsdrupal/drupal-admin-ui/pull/655

PR #629, opened 2019-01-27
created wrapper for empty space

sksanjoo2:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/<TODO>

## Screenshot / UI changes

## Testing instructions

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,12 @@ This is an admin UI for Drupal, built with JavaScript and based on [create-react
 
 ### Docker
 
-Ensure that you currently have nothing running on port 80 or port 3000, for example using ` lsof -i :80 -S`.
+Ensure that you currently have nothing running on port 80 (e.g. a local webserver) or port 3000 (e.g. a local node process). You can run ` lsof -i :80 -S` to see what you currently have running.
+
+Optional: If you're on a Mac and already use yarn locally, may want to move your yarn cache
+folder to a standard location to speed up installs (your local and container can then share the cache).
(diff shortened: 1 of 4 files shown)


martinfrances107:
@sksanjoo2 

Hello welcome to the project.

I have a couple of comments 

a) When filling out the new pull request template, if you add the issue number to the end of the URL then it will bind the PR to the issue.  .. like this.

> Issue
> https://github.com/jsdrupal/drupal-admin-ui/issues/508
> Screenshot / UI changes

b) When I look at the code regarding the use InlineMessage.

here is a line to blog 
https://codeburst.io/using-promises-and-error-boundaries-for-data-loading-within-your-react-redux-app-219f8d3a6d26

shows that the type of error as passed into componentDidCatch can be anything. In this case a promise

```
  componentDidCatch(error, info) {
    // if we have a promise then we can deal with it
    if(error instanceof Promise) {
      // we have a promise so lets update our state
      this.setState({ isLoading: true });
      // once the promise has resolved we can update our state again and grab the data
      error.then((data) => this.setState({ isLoading: false, data }));
    } 
```

looking at our implementation of ErrorBoundary passes the error into InlineMessage message property 


```
  render() {
    if (this.state.error) {
      const pe = new PrettyError();

      return (
        <InlineMessage
          message={pe.render(this.state.error)}
          messageSeverity={MESSAGE_SEVERITY_ERROR}
        />
      );
    }
```

So in short, we do not know in advance the type of message.

Can I ask you to add a conditional statement so that we only wrap strings with a P element.

something like 

```
if (typeof error === 'string' )
  <p>{message}</p>
else
  {message}
```

Anyway I hope this logic makes sense - I hope this helps.

PR #695, opened 2019-04-09
Fix one potential random failure reason

dawehner:
There are quite a lot of potential random failures right now
One example can be seen in https://circleci.com/gh/jsdrupal/drupal-admin-ui/8857#artifacts/containers/0

Maybe checking for the button and not just the snackbar itself fixes some potential race condition,
but who knows.

### DIFF
--- a/packages/admin-ui/src/tests/Nightwatch/Tests/NodeAdd.js
+++ b/packages/admin-ui/src/tests/Nightwatch/Tests/NodeAdd.js
@@ -11,7 +11,7 @@ module.exports = {
     browser
       .relativeURL('/node/add/article')
       .waitForElementVisible('#title')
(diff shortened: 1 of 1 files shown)


packages/admin-ui/src/tests/Nightwatch/Tests/NodeAdd.js
```diff
@@ -11,7 +11,9 @@ module.exports = {
     browser
       .relativeURL('/node/add/article')
       .waitForElementVisible('#title')
-      .waitForElementVisible('[data-nightwatch="restore-content-snackbar"]');
+      .waitForElementVisible(
+        '[data-nightwatch="restore-content-snackbar"] button:first-of-type',
```
dawehner: No idea whether this actually fixes anything ¯\_(ツ)_/¯

PR #701, opened 2019-04-12
Removes using containers for the node packages

justafish:
## Testing instructions

Follow the README instructions and ensure you can run this project


dawehner:
> ERROR: for drupal_admin_ui_drupal  Cannot start service drupal: oci runtime error: container_linux.go:265: starting container process caused "exec: \"sudo supervisord -c /etc/supervisord.conf\": stat sudo supervisord -c /etc/supervisord.conf: no such file or directory"

dawehner:
I've pushed two commits which are required to make this working  (right environment variable and a small fix to the docker container itself)

PR #708, opened 2019-04-16
[#691] First pass at adding a generic <Label> component

eojthebrave:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/691

Add a generic `<Label/>` component for use on forms.

There's not really anything to see yet since the component isn't used anywhere. You can run storybook to see it in action though.

This is probably also not complete as this is just a first pass at figuring out how to create one of these generic form components. It's also my first go at creating a component for this project so feel free to suggest alternative approaches to anything. :)

Example use:

```jsx
<Label htmlFor="username" required={true} htmlAttributes={['data-username' => 'snorfblat']}>Username:</Label>
```

In order to allow someone to pass along generic HTML attributes to the component I've added an `htmlAttributes` prop. With the idea that you populate that with any attributes as key/value pairs, and they get spread to the component. e.g. `<Label htmlFor="username" htmlAttribues={{'data-username': 'snorfblat'}}>Username:</Label>`. And alternative option would be to just spread all non-used props into the `<label>` element. I personally like using `htmlAttributes` as it gives us a property we can document. But open to whatever.

I think whatever we decide here should be the pattern we use for the remaining generic components.

Anyway, here's a start at least.

### DIFF
--- a/packages/components/src/Components/FormElements/FormElements.stories.js
+++ b/packages/components/src/Components/FormElements/FormElements.stories.js
@@ -1,0 +1,19 @@
+import React from 'react';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { storiesOf } from '@storybook/react';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { boolean, object, text } from '@storybook/addon-knobs/react';
+
+import Label from './Label';
+
+storiesOf('FormElements/Label', module).addWithJSX('Default', () => (
+  <Label
+    fieldName="ControlOne"
+    labelFor="asdf"
+    error={boolean('Label: error', false)}
+    required={boolean('Label: required', false)}
+    htmlAttributes={object('Label: htmlAttributes', { 'data-test': 'test' })}
+  >
+    {text('Label: label', 'Field label')}
+  </Label>
+));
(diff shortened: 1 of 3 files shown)


packages/components/src/Components/FormElements/Label.js
```diff
@@ -0,0 +1,83 @@
...
+  children: PropTypes.oneOfType([
+    PropTypes.arrayOf(PropTypes.node),
+    PropTypes.node,
+  ]).isRequired,
+
+  /** true if the label relates to an element with an error. */
+  error: PropTypes.bool,
+
+  /** Any additional HTML properties to add to the label element. */
+  htmlAttributes: PropTypes.objectOf([
```
dawehner: One advantage of this approach is that changing this behaviour is still possible, with a BC layer.

packages/components/src/Components/FormElements/FormElements.stories.js
```diff
@@ -0,0 +1,19 @@
...
+import { storiesOf } from '@storybook/react';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { boolean, object, text } from '@storybook/addon-knobs/react';
+
+import Label from './Label';
+
+storiesOf('FormElements/Label', module).addWithJSX('Default', () => (
+  <Label
+    fieldName="ControlOne"
+    labelFor="asdf"
```
dawehner: I think you changed the prop at some point

```suggestion
    htmlFor="asdf"
```

Issue #705, opened 2019-04-15
Integrate figma into our storybook


dawehner:
https://github.com/hharnisc/storybook-addon-figma allows you to tightly integrate figma with the storybook.

This would allow @ckrina to check the react components for example.

Issue #632, opened 2019-01-28
Storybook: Fix issue with FileUpload


martinfrances107:
There is a issue to create a container for storybook

https://github.com/jsdrupal/drupal-admin-ui/issues/630
https://github.com/jsdrupal/drupal-admin-ui/pull/631

Now that we have a candidate storybook sever, it is clear that we have not been keeping the storybook definitions current. 

I want to fix all the errors with storybook.js files here.

martinfrances107:
ok looking arround the net for similar bugs 

it looks like it is a common error when people try and configure CRA with type script / commonjs

here is on suggestion from 

https://github.com/webpack/webpack/issues/3974

In your tsconfig.json:
Make sure you have this:

    "target": "es2015",
    "module": "es2015",
    "moduleResolution": "node"
I have accidentally put module: "commonjs" and that was causing this error.

In your .babelrc:
Make sure you have set modules to false and loose to true:

"presets": 
[
   [      
      "env",
      {
        "modules": false,
        "loose":  true,
        // ...

iberdinsky-skilld:
on my side setting https://babeljs.io/docs/en/babel-preset-env#modules fixed the problem
[![Image from Gyazo](https://i.gyazo.com/a6b5e2dccfd87387ff7119ed14969c5e.png)](https://gyazo.com/a6b5e2dccfd87387ff7119ed14969c5e)

iberdinsky-skilld:
and this is not only issue with FileUpload.stories. when i deleted that file error was with next component

iberdinsky-skilld:
environment still not really clear for me. so i'm not sure should i create PR here.

martinfrances107:
@iberdinsky-skilld  thank you very much for looking at this... and for providing the solution

I will find time to have a review of your solution this weekend.

What gives me pause for thought is making being sure this compatible with the tsconfig.json files settings given to us by React CRA. and making sure that changing this setting from "auto" (the default) to false does not have any unforeseen consequences down the line. 

For example here is a user - asking for help with tree shaking -- with his reasoning as to when module false causes problems. 

https://stackoverflow.com/questions/47663486/webpack-3-babel-and-tree-shaking-not-working

I am not saying we have the problem .. I just want to do a little investigation.

iberdinsky-skilld:
sure, why not.

iberdinsky-skilld:
seems storybook issue 
https://github.com/storybooks/storybook/issues/4672

PR #710, opened 2019-04-17
implemented figma. added few widgets.

iberdinsky-skilld:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/705

## Screenshot / UI changes

Added figma integration. As [decorator](https://github.com/hharnisc/storybook-addon-figma#or-use-the-decorator-to-put-the-same-design-on-each-story) 

[Image from Gyazo](https://gyazo.com/e405250bedf818e0a5421871b5ba8fce)

Added few components figma integration. In all other existing components stories i wasn't sure about their designs location. 

## Testing instructions

Current implementation of storybook is broken, so this pr will work 
# ONLY
with that small [fix](https://github.com/jsdrupal/drupal-admin-ui/issues/632#issuecomment-459645865)

### DIFF
--- a/packages/admin-ui/.storybook/addons.js
+++ b/packages/admin-ui/.storybook/addons.js
@@ -1,3 +1,4 @@
 import '@storybook/addon-actions/register';
 import 'storybook-addon-jsx/register';
 import '@storybook/addon-knobs/register'
+import 'storybook-addon-figma/register'
--- a/packages/admin-ui/package.json
+++ b/packages/admin-ui/package.json
@@ -80,6 +80,7 @@
     "react-test-renderer": "^16.6.3",
     "redux-saga-test-plan": "^3.6.0",
     "storybook": "^1.0.0",
+    "storybook-addon-figma": "^0.1.0",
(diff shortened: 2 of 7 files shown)


dawehner:
Thank you for working on this!
Note: We'll use the storybook in `packages/components` instead, as the previous version was mostly a prototype.

@iberdinsky-skilld I'm wondering whether @ckrina could split up the figma templates into a reference per component? This would make it nicer.

> Current implementation of storybook is broken, so this pr will work

Could you add a PR for itself?

iberdinsky-skilld:
@dawehner please recheck.
1. I added same steps to components workspace. also added teststory to only one existing button component. 

[Image from Gyazo](https://gyazo.com/d88446c16487a404dba96bdf5d6874da)

2. also it is wierd but i didn't have issues with building storybook in components workspace. so maybe 

> Could you add a PR for itself?

not required anymore?

3. I think it is ok to have all modifiers in one page of figma. Storybook should display all of them aswell:

like here: https://danieldelcore.github.io/trousers/?path=/story/basic--elements-and-modifiers

Issue #733, opened 2019-05-13
react-scripts update to 3.0.1


martinfrances107:
As part of ensuring we use the latest create react app.

We need to bump the major version of react-scripts 

- "react-scripts": "2.1.1",
+"react-scripts": "3.0.1",

A few other changes we forced upon me in order to run  http://localhost:3000/

the eslint version had to be bumped.

I am about to generate to pull request.

martinfrances107:
I think the  associated PR is ready for review ...

Here is a summary of the changes 

1) My motivation was to upgrade to the latest create-react-app. There are minor incremental improvements to typescript support.

2) Practically the means a bump in the react-scripts version number.
2a The new react-scripts reports errors unless the eslint version number is bumped. 
2b) The updated linter reports deprecated warning associate with the history package [ which was updated and our code updated] 
2c) The updated linter demands some proptypes be locked/down for tests to pass.

dawehner:
@martinfrances107 thank you for your great PR! I've merged it :0

PR #734, opened 2019-05-13
#733 react-scripts: update to 3.0.1

martinfrances107:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/733

## Screenshot / UI changes

## Testing instructions

I want to upgrade to the latest create-react-app 

This means bumping the package versions of 

react-scripts
eslint


packages/components/src/Components/FormElements/InputCheckbox/InputCheckbox.js
```diff
@@ -142,25 +142,33 @@ const InputCheckbox = props => {
 };
 
 InputCheckbox.propTypes = {
+  checked: PropTypes.bool,
```
dawehner: In general, try to make these changes in a separate PR, they aren't really related with updating create-react-app :)

martinfrances107:
Thanks for looking at this

Ah the test would not pass unless changes to the prop types were made

Like a game of dominoes 

changing the react-scripts triggered a change in linter. etc etc.
 
Maybe the changes to default settings ... could have been done in a separate issue.

dawehner:
I totally get this. The changes of the proptypes you can though still do in a separate PR and make it a dependency of some other one.

martinfrances107:
Ah I see what you mean ... I should have spawned 3ish child issues...I will adjust my approach in future.

dawehner:
Thank you Martin.

I always try to assume that the person reviewing has minimal time, so I try
to make their life easier. I make mistakes all the time, but it's a good
thing to keep in mind.

Issue #738, opened 2019-05-16
Keyboard experience requires twice the effort in Menu


anevins12:
Hey I'm new here,

I found that the menu which appears to the side of content takes twice as many tabs to navigate through it than it should. This is because the `<li>` of each menu link has a `tabindex` of `0`, which puts it into the keyboard navigation. 

May I ask why the `<li>` elements have this tabindex? If the reason is that they need to have focus in other environments maybe we can change the `0` to `-1`. That removes them from the natural keyboard navigation but sitll makes them programmatically focusable.

I've attached a screenshot that outlines 3 of these links in the menu but this issue applies to all `<li>`s in the menu that surround link.

Happy to look into this issue after discussion.

![download (39)](https://user-images.githubusercontent.com/2318561/57859863-aaea1c00-77eb-11e9-85e2-52399170ab7d.png)

anevins12:
This looks like it's the default config coming form Material-UI Core:

```
ButtonBase.defaultProps = {
  centerRipple: false,
  component: 'button',
  disableRipple: false,
  disableTouchRipple: false,
  focusRipple: false,
  tabIndex: '0',
  type: 'button'
};
```
_drupal-admin-ui/node_modules/@material-ui/core/ButtonBase/ButtonBase.js_

Looks like we can override that.

anevins12:
Note to self: Try to stop the `<li>` from rendering the button component: https://github.com/jsdrupal/drupal-admin-ui/blob/master/packages/admin-ui/src/components/06_wrappers/Default/Default.js

anevins12:
Okay I kinda get why the Button component was used on the `<li>`. It's for styling purposes. Looks like we'll have to add some more CSS to the `<li>` instead - Expect CSS on the PR.

anevins12:
As we'll be moving away from the design and CSS of Material UI in future, I'm intentionally leaving the CSS out of this Merge Request.

martinfrances107:
@anevins12 

Thank you for pointing this out ... 

When I looked at your PR and saw that you removed the button property... it triggered something in the back of my subconscious.

That code is brittle .. well it is difficult in the context of material-ui to get the components into a satisfactory form.

I expect that as you are using a library that all the 'ally' issues  would be taken into consideration and that all best practise would be baked into the patterns by default.

This is not the case and it is just finicky to get right.

You would mind if I expanded the drive behind your issue to fix a related problem?

If you run 

[...]

PR #739, opened 2019-05-17
738: Removed button component from elements in the Drawer navigation

anevins12:
## Issue
https://github.com/jsdrupal/drupal-admin-ui/issues/738 

## Screenshot / UI changes
The interaction events on hover, click and focus no longer apply to the `<li>` and so the UI looks the same as it does in its default state. However the outline of the focus indicator persists (yay).

download (40)

## Testing instructions
- Check if the interface works as it did before;
- With a keyboard and with the 'tab' key, navigate through the Drawer navigation and check if you need to press the 'tab' key once to get to the next Drawer item.

### DIFF
--- a/packages/admin-ui/src/components/06_wrappers/Default/Default.js
+++ b/packages/admin-ui/src/components/06_wrappers/Default/Default.js
@@ -90,7 +90,6 @@ class Default extends React.Component {
             <ListItem
               key={menuLink.url.replace(/\//g, '-')}
               component="li"
-              button
             >
               <Link to={menuLink.url} className={styles.menuLink} role="button">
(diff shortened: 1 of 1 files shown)


dawehner:
@anevins12 If you run `yarn run eslint` locally it should fix the eslint problem automatically for you :) This should make the tests passing.

anevins12:
> @anevins12 If you run `yarn run eslint` locally it should fix the eslint problem automatically for you :) This should make the tests passing.

I'm just wondering how you read the error log that indicated the linting issue. At the moment I'm finding it difficult identifying the errors from the failed pipelines.

Issue #752, opened 2019-05-29
Duplicate  parse and parserOptions in .eslintrc.json


martinfrances107:
While converting drupal/admin-ui-utilities into typescript I noticed this little duplication error.

My attempt to convert code section into typescript -- by their nature become hard to review.

So I am splitting this issue off into a small contained task.

Here is the code form eslintrc.json which show where to duplication is 

```
{
  "extends": [
    "airbnb",
    "plugin:prettier/recommended",
    "plugin:jsx-a11y/recommended"
  ],
  "parser": "typescript-eslint-parser",
  "parserOptions": {
    "jsx": true
  },
  "parser": "babel-eslint",
  "parserOptions": {
      "ecmaVersion": 6,
      "ecmaFeatures": {
          "jsx": true
      }
  },
```

I am about to create a PR

PR #753, opened 2019-05-29
eslintrc.json: Removed duplicate parserOptions

martinfrances107:
## Issue

https://github.com/jsdrupal/drupal-admin-ui/issues/752

.tslintrc.json incorrectly has the following duplicate keys parser and parserOptions

Removed the keys which are being overridden.


martinfrances107:
thanks

